PR #24, opened 2019-03-12
Basic service impl

moshababo:
Continuing previous commit. 
This is WIP, but need to be merged before the following gRPC PR. 

Documentation & tests will be added after API will get stabilized.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -1,7 +1,8 @@
 package service
 
 import (
-	"fmt"
+	"errors"
+	"github.com/spacemeshos/poet-ref/shared"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


service/service_test.go
```diff
@@ -34,23 +36,62 @@ func TestNewService(t *testing.T) {
...
 
 	leafIndices := []uint64{uint64(size - 1)}
 	leaves := [][]byte{lastCommitment}
-	valid, err := merkle.ValidatePartialTree(leafIndices, leaves, proof, round.treeRoot, merkle.GetSha256Parent)
+	valid, err := merkle.ValidatePartialTree(leafIndices, leaves, proof, round.merkleRoot, merkle.GetSha256Parent)
 	req.NoError(err)
 	req.True(valid)
 }
+
+//func TestService_Info(t *testing.T) {
```
antonlerner: if you wish to keep this code but not run it yet I'd recommend using t.Skip()

service/service.go
```diff
@@ -42,52 +67,103 @@ func NewService(cfg *Config) (*Service, error) {
...
 }
 
-func (s *Service) SubmitCommitment(c []byte) (*submitResponse, error) {
-	r := s.activeRound
+func (s *Service) Info() *InfoResponse {
+	res := new(InfoResponse)
+	res.openRoundId = s.openRound.id
+
+	ids := make([]int, 0, len(s.executingRounds))
+	for id := range s.executingRounds {
```
antonlerner, line 111: must be interlocked with NewService

service/service.go
```diff
@@ -42,52 +67,103 @@ func NewService(cfg *Config) (*Service, error) {
...
+
+func (s *Service) SubmitCommitment(c []byte) (*SubmitCommitmentResponse, error) {
+	r := s.openRound
 	err := r.submitCommitment(c)
 	if err != nil {
 		return nil, err
 	}
 
-	res := new(submitResponse)
+	res := new(SubmitCommitmentResponse)
```
antonlerner, line 158: no actual need for new, you can also refactor the two lines to ine using `res := SubmitCommitmentResponse {roundId : r.id}`

service/service.go
moshababo: Sorry, I didn't understand what you mean.

service/service.go
moshababo: Do you mean `&SubmitCommitmentResponse {roundId : r.id}`?
If so, since they are equivalent, you mean that that’s the preferred style?
I'm asking because I used `new` in all other pointer constructs as well.

service/service.go
antonlerner: `executingRounds` is accessed here and in another go routine, please use a mutex when reading from same memory concurrently.

service/service.go
antonlerner: Yes, using `new` keyword is not common in go apps, but it is not important to refactor it. Just wanted to mention the alternative that is also more compact

service/service.go
moshababo: Yes, i'll left a TODO for this in the other goroutine. I'll use the same mutex which will be applied there.

PR #26, opened 2019-03-19
Service test

moshababo:


service/round.go
```diff
@@ -89,19 +90,19 @@ func (r *round) execute() error {
 func (r *round) membershipProof(c []byte) ([][]byte, error) {
 	// TODO(moshababo): change this temp inefficient implementation
 	ci := -1
-	for i, commitment := range r.commitments {
-		if bytes.Equal(c, commitment) {
+	for i, commit := range r.commits {
+		if bytes.Equal(c, commit) {
```
antonlerner, line 94: style: single letters are usually reserver for function receiver, IMO describing what c means won't hurt and maybe even make code more readable

PR #28, opened 2019-03-27
Proofs RPC long-polling

moshababo:
Implemented long-polling support for `GetMembershipProof` & `GetProof` RPC calls. I didn't use gRPC stream (server side push) because there's only one update, so I thought that long-polling is more appropriate. 

Server timeouts are not handled at the moment, since they are expected to be very long anyway.

### DIFF
--- a/service/round.go
+++ b/service/round.go
@@ -11,7 +11,8 @@ import (
 )
 
 type round struct {
-	cfg          *Config
+	cfg *Config
+
 	Id           int
 	opened       time.Time
 	executeStart time.Time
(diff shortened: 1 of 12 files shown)


service/round.go
```diff
@@ -87,7 +88,17 @@ func (r *round) execute() error {
 	return nil
 }
 
-func (r *round) membershipProof(c []byte) ([][]byte, error) {
+func (r *round) membershipProof(c []byte, wait bool) ([][]byte, error) {
+	if wait {
+		<-r.closedChan
+	} else {
+		select {
+		case <-r.closedChan:
```
antonlerner, line 96: why read from channel here if you have a default statement?

moshababo: to check whether `closedChan` was closed. if so, then it will proceed, instead of the default case, which basically checks if it wasn't closed.

PR #40, opened 2019-04-21
Add travis file

beckmani:


Makefile
```diff
@@ -0,0 +1,26 @@
+BINARY := poet-ref
+DOCKER_IMAGE_REPO := poet-ref
```
moshababo, line 2: Can use `poet` instead of `poet-ref`. We'll change the project name soon, in the meanwhile we can specify the output name.

Issue #37, opened 2019-04-16
Use the new merkle-tree implementation


moshababo:
The number of leafs to prove depends on the security parameter constant (`T`).
[merkle-tree](https://github.com/spacemeshos/merkle-tree) can be used to create more efficient proof of multiple leafs, specially for the verifier which currently use a key-value cache to detect repeating nodes in the tree.

moshababo:
Closed by #49.

PR #49, opened 2019-05-01
New poet

noamnelke:
Still a draft. Todo:

- [x] Connect to outside (grpc)
- [x] Remove old code (incl. some renaming and other cleanup)
- [x] More tests and benchmarks
- [x] After merging https://github.com/spacemeshos/merkle-tree/pull/7 - point dependency to develop (currently points to working branch)

Closing #37.


prover/newprover.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/spacemeshos/merkle-tree"
+	"github.com/spacemeshos/merkle-tree/cache"
+	"github.com/spacemeshos/poet/shared"
+)
+
+func GetProof(challenge shared.NewChallenge, leafCount uint64, securityParam uint8) (shared.MerkleProof, error) {
+	treeCache := cache.NewWriter(
+		cache.Combine(
+			cache.SpecificLayersPolicy(map[uint]bool{0: true}),
+			cache.MinHeightPolicy(11)),
```
zalmen: make the min height configurable (or at least declared in a global var)

prover/newprover.go
```diff
@@ -0,0 +1,37 @@
...
+		cache.MakeSliceReadWriterFactory())
+	tree := merkle.NewTreeBuilder().WithHashFunc(challenge.MerkleHashFunc()).WithCacheWriter(treeCache).Build()
```
zalmen: Why do you get the hash function from outside? How do you know that it is the same one that was used to compute the tree?

prover/newprover.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+	}
+	root := tree.Root()
+
+	cacheReader, err := treeCache.GetReader()
+	if err != nil {
+		return shared.MerkleProof{}, err
+	}
+	provenLeafIndices := shared.FiatShamir(root, leafCount, securityParam)
+	_, provenLeaves, proofNodes, err := merkle.GenerateProof(provenLeafIndices, cacheReader)
```
zalmen: You should handle `err != nil`

verifier/newverifier.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/spacemeshos/poet/shared"
+	"sort"
+)
+
+func Validate(proof shared.MerkleProof, challenge shared.NewChallenge, leafCount uint64, securityParam uint8) error {
+	if int(securityParam) != len(proof.ProvenLeaves) {
+		return fmt.Errorf("number of proven leaves (%d) must be equal to security param (%d)",
+			len(proof.ProvenLeaves), securityParam)
+	}
+	provenLeafIndices := asSortedSlice(shared.FiatShamir(proof.Root, leafCount, securityParam))
```
zalmen: The pattern of calling FiatShamir and then sorting the result repeats itself. Consider wrapping it in a single function (will also prevent bugs where people forget to sort the result)

verifier/newverifier.go
```diff
@@ -0,0 +1,45 @@
...
+		return fmt.Errorf("error while validating merkle proof: %v", err)
+	}
+	if !valid {
+		return fmt.Errorf("merkle proof not valid")
+	}
+
+	for id, label := range proof.ProvenLeaves {
+		expectedLabel := shared.MakeLabel(challenge.LabelHashFunc(), provenLeafIndices[id], parkingSnapshots[id])
+		if !bytes.Equal(expectedLabel, label) {
+			return fmt.Errorf("label at index %d incorrect", id)
```
zalmen: add a log print (error) with the expected and actual results

verifier/newverifier.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+	}
+
+	return nil
+}
+
+func asSortedSlice(s map[uint64]bool) []uint64 {
+	var ret []uint64
+	for key, value := range s {
```
zalmen: Why do you pass it as a dictionary to begin with? Isn't it better if the FiatShamir method will return a set instead of a map?

prover/newprover.go
noamnelke: 😱of course

prover/newprover.go
noamnelke: The hash function is derived from the challenge (the challenge is used as salt). The verifier should use the expected challenge (the root if the commitment tree) and see if the root is identical.

If you're wondering why I'm using an interface for the challenge rather than pass the real object, it's because the existing code supports different hash functions (sha256 / scrypt) and I know this is not finalized yet, so I wanted to support easily benchmarking different functions.

Now that I'm thinking about it some more - I may as well hard code it and benchmark a different branch for each hash function, since it's not something we'll want to make configurable eventually.

verifier/newverifier.go
noamnelke: if all usages of this function would require a sorted list, I'd agree. However, some need the raw output (`map[uint64]bool`) which is why I've created a function to transform it into a sorted slice.

Since the data types are different, there's no risk of someone forgetting to run the wrapping function.

verifier/newverifier.go
noamnelke: Note that `merkle.GenerateProof()` expects it that way (`map[uint64]bool`). I assume that by "set" you mean a sorted slice?

verifier/newverifier.go
zalmen: All usages of the Fiat Shamir method are also sorted. I see no real reason to use a map and not a slice

prover/diskreadwriter.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"bufio"
+	"github.com/spacemeshos/merkle-tree"
+	"io"
+	"os"
+)
+
+const OwnerReadWrite = 0600
+
+func NewDiskReadWriter(filename string) *DiskReadWriter {
```
zalmen: Consider moving it to the merkle repo (and use it both here and in the PoST code)

zalmen: Also, plz document all exported methods

prover/diskreadwriter.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (rw *DiskReadWriter) Width() uint64 {
+	err := rw.b.Flush() // TODO(noamnelke): add flush method to interface and call when switching to read mode
+	if err != nil {
+		log.Error("failed to flush disk writer: %v", err)
+		return 0
+	}
+	info, err := rw.f.Stat()
+	if err != nil {
+		log.Error("failed to get stats for disk reader: %v", err)
```
zalmen: you need to return the error, it is expected especially due to the signature of the other methods

prover/readwritermetafactory.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/spacemeshos/merkle-tree/cache"
+	"os"
+)
+
+type ReadWriterMetaFactory struct {
+	minMemoryLayer uint
+	filesCreated   []string
+}
+
+func NewReadWriterMetaFactory(minMemoryLayer uint) *ReadWriterMetaFactory {
```
zalmen, line 20: Add documentation

prover/readwritermetafactory.go
```diff
@@ -0,0 +1,44 @@
...
+	return &ReadWriterMetaFactory{minMemoryLayer: minMemoryLayer}
+}
+
+func (mf *ReadWriterMetaFactory) GetFactory() cache.LayerFactory {
+	return func(layerHeight uint) cache.LayerReadWriter {
+		if layerHeight < mf.minMemoryLayer {
+			fileName := makeFileName(layerHeight)
+			readWriter := NewDiskReadWriter(fileName)
+			mf.filesCreated = append(mf.filesCreated, fileName)
```
zalmen: it is possible that the class user already asked for this layer and you already have it in the cache. Consider replacing the slice with a map.

zalmen (changes requested):

prover/diskreadwriter.go
noamnelke: I was just lazy 😉 done now

zalmen (approved):

PR #50, opened 2019-05-23
Integrate gossip (via node)

noamnelke:
This can't be tested yet, since there's a "circular dependency" between the node and the PoET server (each needs to know the other's IP address when starting).

A temporary solution is to bring up both machines with a static IP so we can point the other to it. Later we should implement an endpoint on the node to set the PoET address (which we'd probably need for the app at some point anyway).


zalmen:
Commenting on your comment above regarding the circular dependency -
A simpler way to handle this is to do the initialization in two steps - the first step is to start the service and the second step will be calling an initialization call with the dependencies needed.

broadcaster/pb/api.proto
```diff
@@ -0,0 +1,119 @@
+syntax = "proto3";
```
zalmen, line 1: We should have a dedicated repo for shared resources, such as this, instead of copying it over and over

broadcaster/pb/api.pb.go
```diff
@@ -0,0 +1,876 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// source: api.proto
+
+package pb
```
zalmen, line 4: Why do we need to have auto-generated code in the repo?

service/service.go
```diff
@@ -114,6 +142,29 @@ func NewService(cfg *Config) (*Service, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get poet proof for round %d: %v", r.Id, err)
+	}
+	proofMessage := PoetProofMessage{
+		GossipPoetProof: GossipPoetProof{
+			MerkleProof: *r.nip,
+			Members:     r.challenges,
+			LeafCount:   uint64(1) << poetProof.N,
+		},
+		PoetId:    nil,
```
zalmen: why nil? and what about signing the msg?

service/service.go
```diff
@@ -104,6 +126,12 @@ func NewService(cfg *Config) (*Service, error) {
 					log.Error(err)
 				}
 
+				if msg, err := serializeProofMsg(r); err != nil {
+					log.Error(err)
+				} else if err := broadcaster.BroadcastProof(msg); err != nil {
```
zalmen: consider calling `BroadcastProof` in a go routine, no reason to wait for the result (if you decide to change it, remember to add a log print in case of error inside the method, instead of here)

zalmen (changes requested):

broadcaster/pb/api.proto
noamnelke: I agree. I discussed this with @moshababo and he says it's common practice to copy-paste proto files between repos, so I chose the less-work way for now. I do agree this should be refactored, though.

broadcaster/pb/api.pb.go
noamnelke: All the protobuf-generated files in the PoET repo are checked-in to git. There are pros and cons to this. I'm actually for including them as it reduces headaches around remembering to regenerate when someone else pushes a change and the generated code isn't platform-dependant or anything.

I don't mind changing this, but then we should be consistent and git-ignore **all** protobuf-generated files.

service/service.go
noamnelke: PoET identities aren't implemented yet and I don't think they should be part of this PR. If you disagree, lmk. (this should answer both your questions)

noamnelke:
@zalmen about the circular dependency - yes, I agree, I just think this should be on the node, not the PoET server.

The PoET server will **always** need a node attached to broadcast proofs, while the node needs a PoET server only for mining. The user should select a PoET server when they enable mining. I think we should implement an endpoint on the node for enabling mining which gets as input the PoET server address and number of "space units" to initialize.

I think this would be very easy to implement (starting the ATX builder loop when this endpoint is called instead of when the node starts). The PoET server address should then be stored in the NIPST builder singleton instead of the config.

y0sher:
the log changes look good to me.

zalmen (approved):

PR #57, opened 2019-07-08
Process cleanup

moshababo:
When creating a new test harness instance, verify no other PoET service is listening on the same port. If such exist, kill it.


integration/harness.go
```diff
@@ -87,3 +98,41 @@ func baseDir() (string, error) {
...
+	if err != nil {
+		return err
+	}
+
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" {
+		args := fmt.Sprintf("(Get-NetTCPConnection -LocalPort %s).OwningProcess -Force", addr.Port)
+		cmd = exec.Command("Stop-Process", "-Id", args)
+	} else {
+		args := fmt.Sprintf("lsof -i tcp:%d | grep LISTEN | awk '{print $2}' | xargs kill -9", addr.Port)
```
antonlerner, line 122: this is ubuntu specific commands, maybe ps is better? or just use go os

moshababo: What do you mean by use go os? to inspect the platform?

antonlerner (approved):

PR #59, opened 2019-07-15
Remove support for previous rounds state

moshababo:
It is no longer necessary after the service was changed to actively broadcast it.


server.go
```diff
@@ -60,7 +60,7 @@ func startServer() error {
 		grpcServer = grpc.NewServer(options...)
 
 		api.RegisterPoetServer(grpcServer, rpcServer)
-		proxyRegstr = append(proxyRegstr, api.RegisterPoetHandlerFromEndpoint)
+		//proxyRegstr = append(proxyRegstr, api.RegisterPoetHandlerFromEndpoint)
```
noamnelke: you can re-enable this now, I think

noamnelke (approved):

PR #60, opened 2019-07-16
Return random poet id

noamnelke:


rpc/api/api.proto
```diff
@@ -25,6 +25,7 @@ message SubmitRequest {
 
 message SubmitResponse {
     int32 roundId = 1;
+    bytes poetId = 2; // TODO: remove this once we have persistent PoET identities that the node knows in advance
```
moshababo: I found the name a bit confusing, because "poet" refers to a proof. How about `serviceId`?
I know it's already implemented in other places, so you can leave it as is if you prefer.

rpc/rpcserver.go
```diff
@@ -23,13 +23,14 @@ func NewRPCServer(service *service.Service) *rpcServer {
...
-	res, err := r.s.Submit(in.Challenge)
+	round, err := r.s.Submit(in.Challenge)
 	if err != nil {
 		return nil, err
 	}
 
 	out := new(api.SubmitResponse)
-	out.RoundId = int32(res.Id)
+	out.RoundId = int32(round.Id)
+	out.PoetId = round.PoetId[:]
```
moshababo: You should add it to `GetInfo` as well, although it's not used much yet.

moshababo (approved):

PR #62, opened 2019-07-29
fixed bad import for the log package

zalmen:


moshababo (changes requested):
`spacemeshos/go-spacemesh` dependency can be removed from `go.mod`.

noamnelke:
What he ☝️ said, but specifically, run `go mod tidy`

moshababo (approved):

PR #72, opened 2019-10-28
Adding mock node to poet testing

AmitShaul:
WIP.


integration/server.go
```diff
@@ -36,11 +36,12 @@ func DefaultConfig() (*ServerConfig, error) {
...
-		rpcListen: "127.0.0.1:18550",
-		baseDir:   baseDir,
-		dataDir:   filepath.Join(baseDir, "data"),
-		exe:       poetPath,
+		logLevel:  		"debug",
+		rpcListen: 		"127.0.0.1:18550",
+		baseDir:   		baseDir,
+		dataDir:   		filepath.Join(baseDir, "data"),
+		exe:       		poetPath,
+		NodeAddress: 	"localhost:9091",
```
noamnelke, line 44: You should only use tabs for indentation, not spacing within a line. Instead of fixing the indentation yourself, you can use <kbd>⌘ command</kbd>+<kbd>⌥ option</kbd>+<kbd>L</kbd> for the IDE to do it for you.

If you run `go fmt ./...` it will fix all these issues for you (you can configure the IDE to do it on every save).

PR #83, opened 2019-12-14
Support multiple gateway nodes

moshababo:
#### Functionality

* Allow connecting to multiple gateway nodes on server start. **All** connections must succeed in order to proceed.
* Broadcast proofs to all connected nodes. If **one** request succeeded, the proof is considered as broadcasted. Otherwise the round will remain in "unbroadcasted" state, same as if the server crashed (after executing and before broadcasting). This state is settled only by the crash recovery, so we should probably add retry mechanism as well @barakshani @noamnelke. 

#### API changes
* `NodeAddress` changed to `[]GatewayAddresses`
* `DisableBroadcast` was added (instead of the `NO_BROADCAST` string detection)


barakshani:
How does connection to the poet actually happen?
Do we only specify the amount of gateway nodes it needs, and the poet go through a discovery process, or do we specify particular nodes (like bootstrap node today) that need to connect to it?
What happens if one node does not succeed to establish a connection?

moshababo:
> How does connection to the poet actually happen?

This PR isn’t about nodes connection to the poet, but rather the poet connection to gateway nodes (for broadcasting proofs).

> Do we only specify the amount of gateway nodes it needs, and the poet go through a discovery process, or do we specify particular nodes (like bootstrap node today) that need to connect to it?

There’s no discovery - the poet isn’t directly integrated with Spacemesh p2p layer. You need to provide **gRPC** endpoints of the desired gateway nodes.

> What happens if one node does not succeed to establish a connection?

If one connection didn’t succeed, the service won’t start.

moshababo:
I didn't add mock-based unit tests because i'm quite sure we will tweak this functionality later on.

broadcaster/broadcaster.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// Parse and concatenate potential errors.
+	var numErrors int
+	var retErr error
+	for i := range b.clients {
+		err := errs[i]
+		res := responses[i]
+		if err != nil {
+			err = fmt.Errorf("failed to broadcast via gateway node at \"%v\" after %v: %v", b.connections[i].Target(), elapsed, err)
+		} else if res.Value != "ok" {
```
noamnelke, line 147: Minor, style, ignore if you disagree:
```suggestion
		} else if res := responses[i]; res.Value != "ok" {
```
Instead of setting `res` in advance, even if there's an `err`.

broadcaster/broadcaster.go
```diff
@@ -0,0 +1,162 @@
...
+		if err != nil {
+			err := fmt.Errorf("failed to connect to Spacemesh gateway node at \"%v\": %v", gatewayAddresses[i], err)
+			if retErr == nil {
+				retErr = err
+			} else {
+				retErr = fmt.Errorf("%v | %v", retErr, err)
+			}
+		}
+	}
+	if retErr != nil {
```
noamnelke: I think it's fairly easy to set a threshold of required connections as a config, and check for it here, instead of any failure killing the startup (by adding an err counter).

You can then ensure it's `>=` on [line 94](https://github.com/spacemeshos/poet/pull/83/files#diff-88cefb21ab4fec8783af3c062fcf4d07R94-R96).

noamnelke (changes requested):
I think we should set a minimum successful connections threshold instead of requiring that all connections succeed.

noamnelke:
Actually, I also think we need a threshold for minimum successful transmissions.

broadcaster/broadcaster.go
moshababo: IMO this optimization produces a slightly less readable code, so I'd rather leave it.

noamnelke (approved):
:shipit:

Issue #84, opened 2019-12-29
Update gateway after service already started


moshababo:
Add support to update PoET service gateway nodes after it already started. 
At the moment one must restart the server in order to do this.

PR #86, opened 2019-12-29
Update gateway after service already started

moshababo:
Closing https://github.com/spacemeshos/poet/issues/84.

### Changes

Introducing new RPC method: `UpdateGateway`, which has a similar request message as `Start`, but can be used after the service already started. 

## Open issue

Same like `Start`, this new RPC method is admin-only. We didn't introduce authentication to `Start` because it wasn't critical enough, since it was supposed to be called quickly and adversarial consecutive calls would fail anyway. But with `UpdateGateway` it isn't the case, so we need to protect it. 

Simplest option would be to add a configurable password to the server (with a default value which need to be changed on production), and add a password field to the admin-only requests to match it (this will also require to enable TLS for the gRPC connection). 

Any thoughts? @noamnelke @y0sher


noamnelke (approved):
LGTM.

Regarding password protection: I don't like the idea of adding a PW mechanism ourselves. There are many things to consider for such a sensitive service (although it might not be critical until mainnet), e.g. what if someone tries to bruteforce the PW? So we need rate limiting or something. We could probably come up with more possible attacks to protect against.

I think we should take a different approach: we should switch from gRPC to REST regardless of security, but doing so will enable us to put an API Gateway between the internet and the actual PoET server. The gateway can relay only a subset of endpoints to the server, which excludes administrative commands. To run administrative commands we'd need to talk to the server directly which should only be possible from within our VPC, so to call it from our machine we'd need to tunnel through some instance. Under this model, if someone can access the admin functions it means they have terminal access to our AWS instances and then we're screwed anyway.

noamnelke:
cc: @ilans ☝️ 

Please confirm that I'm not talking nonsense and that this would be easy to configure.

moshababo:
We'll want to have the production server sitting behind a gateway anyway, so I tend to agree with you.

@ilans please advise.

noamnelke:
I think you should just merge and we'll figure it out later

Issue #87, opened 2019-12-29
Add retries to broadcast failure


moshababo:
When round broadcast is failing, the round stays in "unbroadcasted" state, and only the recovery mechanism can initiate another attempt to broadcast. This is not enough, and we should introduce delayed retries after the initial failure.

PR #88, opened 2019-12-30
Simple retry mechanism for proof broadcasting

moshababo:
Closing #87.

Introducing retry mechanism for proof broadcasting. 

### Current default values
* BroadcastRetriesInterval: 5min
* BroadcastNumRetries: 100

### DIFF
--- a/shared/shared.go
+++ b/shared/shared.go
@@ -56,6 +56,7 @@ func Retry(retryable func() error, numRetries int, interval time.Duration, logge
 	}
 
 	if numRetries < 1 {
+		logger(fmt.Sprintf("%v", err))
 		return errors.New("number of retries exceeded")
 	}
 
--- a/shared/shared.go
+++ b/shared/shared.go
@@ -2,8 +2,11 @@ package shared
 
 import (
 	"encoding/binary"
+	"errors"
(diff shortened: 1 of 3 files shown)


shared/shared.go
```diff
@@ -45,3 +48,20 @@ type MerkleProof struct {
...
+
+func Retry(retryable func() error, numRetries int, interval time.Duration, logger func(msg string)) error {
+	err := retryable()
+	if err == nil {
+		return nil
+	}
+
+	if numRetries < 1 {
+		logger(err.Error())
+		return errors.New("number of retries exceeded")
```
noamnelke, line 60: I prefer to print a single message, so instead of printing the error and then returning a static message, I'd return:
```golang
return fmt.Errorf("number of retries exceeded: %v", err)
```

noamnelke (approved):
I had one comment that's more a matter of style, so up to you.

shared/shared.go
moshababo: Errors returned from the function calls are logged and not aggregated, to report them immediately. The returned error is only about the overall retries exhaustion. This is why I didn't combine it with the returned error as you suggested.

PR #103, opened 2020-09-28
Print stderr from go build when it fails

lrettig:
Setting up a poet harness from go-spacemesh was failing for me due to an environment issue (two conflicting versions of go). `go build` here was failing silently. This prints the error.


moshababo (approved):
Looks good!

PR #108, opened 2021-08-30
adding `shutdown` grpc endpoint to the poet

AmitShaul:
## description
adding a new grpc endpoint and client for shutting down the poet remotely.
shutdown will only be available if `debug` is added to build tags when building poet binary in order to prevent malicious use of this sensitive functionality.

this PR is part of the state loading process for automation full node tests.
https://github.com/spacemeshos/go-spacemesh/issues/2619

## TODO
- [x] add a `GRPC` client to the PoETs (added under `cmd -> grpc_shutdown`)
- [x] add a `runner` wrapping script for the PoET to work around `kubernetes` high availability behaviour
- [x] use `build flags`
- [ ] proper shutdown functionality - so far we are using `Panic` in order to quit


dshulyak (dismissed):
this service should be started on a non-public endpoint, otherwise, anyone can shutdown poet

cmd/grpc_shutdown/main.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	c := debug.NewDebugClient(conn)
+	res, err := c.Shutdown(context.Background(), &debug.ShutdownRequest{})
+
+	if err != nil {
+		fmt.Printf("Failed to get account details: %s", err)
+	}
+	fmt.Sprintln("result: %v", res)
+}
```
dshulyak, line 26: this can be replaced with grpcurl, an example:

```
grpcurl -plaintext localhost:6000 spacemesh.v1.NodeService.Status
```

Dockerfile
```diff
@@ -18,8 +18,26 @@ FROM build_base AS server_builder
 COPY . .
 
 # And compile the project
+RUN go build -tags debug
```
y0sher, line 21: I don't think it's right to always build with this tag

PR #109, opened 2021-10-10
config and server have their own packages and can be used by a library

y0sher:
Just refactored the functions as exported functions of a package so we can include them in other projects (mainly for testing..).
I believe this structure is more readable as well.


config/config.go
```diff
@@ -149,12 +141,21 @@ func loadConfig() (*config, error) {
...
+		log.Warning("%v", configFileError)
+	}
+
+	// TODO: move out of here
 	// Finally, parse the remaining command line options again to ensure
 	// they take precedence.
-	if _, err := flags.Parse(&cfg); err != nil {
-		return nil, err
-	}
+	return ParseFlags(cfg)
```
fasmat: This imho should just return the config and not call `ParseFlags` (since it is already called in `poet.go` before this function is called.

The comment above suggests that this is intentional, but I don't think it is?

config/config.go
```diff
@@ -149,12 +141,21 @@ func loadConfig() (*config, error) {
...
 	}
 
+	// Warn about missing Config file only after all other configuration is
+	// done.  This prevents the warning on help messages and invalid
+	// options.  Note this should go directly before the return.
+	if configFileError != nil {
+		log.Warning("%v", configFileError)
+	}
+
+	// TODO: move out of here
```
fasmat: What is this TODO about?

config/config.go
```diff
@@ -149,12 +141,21 @@ func loadConfig() (*config, error) {
 		configFileError = err
 	}
 
+	// Warn about missing Config file only after all other configuration is
+	// done.  This prevents the warning on help messages and invalid
+	// options.  Note this should go directly before the return.
+	if configFileError != nil {
+		log.Warning("%v", configFileError)
```
fasmat, line 151: Imho this error should be returned instead of logged, so it can be handled.

config/config.go
```diff
@@ -149,12 +141,21 @@ func loadConfig() (*config, error) {
...
 	// Finally, parse the remaining command line options again to ensure
 	// they take precedence.
-	if _, err := flags.Parse(&cfg); err != nil {
-		return nil, err
-	}
+	return ParseFlags(cfg)
+}
 
+// SetupConfig initializes filesystem and network infrastructure.
+func SetupConfig(cfg *Config) (*Config, error) {
```
fasmat, line 158: The logging in this function is now incorrect. It logs as `loadConfig` instead of as `SetupConfig`.

Also the error handling in general is outdated; instead of casting the `err` variable, an `errors.As` might be a more modern approach.

poet.go
```diff
@@ -11,24 +11,36 @@ import (
...
+	"github.com/spacemeshos/poet/config"
+	"github.com/spacemeshos/poet/server"
+)
 
 // poetMain is the true entry point for poet. This function is required since
 // defers created in the top-level scope of a main method aren't executed if
 // os.Exit() is called.
 func poetMain() error {
 	// Use all processor cores.
 	runtime.GOMAXPROCS(runtime.NumCPU())
```
fasmat: From godoc:

> `GOMAXPROCS` sets the maximum number of CPUs that can be executing simultaneously and returns the previous setting. It defaults to the value of `runtime.NumCPU`

So this is function call does nothing 😅

poet.go
```diff
@@ -11,24 +11,36 @@ import (
...
 
-var cfg *config
+	"github.com/spacemeshos/poet/config"
+	"github.com/spacemeshos/poet/server"
+)
 
 // poetMain is the true entry point for poet. This function is required since
 // defers created in the top-level scope of a main method aren't executed if
 // os.Exit() is called.
 func poetMain() error {
```
fasmat, line 22: I think it is weird that `poetMain()` is in the `config` package.

EDIT: ignore this comment, my mistake.

fasmat (commented):
There is some confusing stuff going on the config parsing and this quick fix feels like it makes it more confusing.

poet.go
poszu: It's in the `main` package.

fasmat (approved):

config/config.go
poszu: No, please read the comments in the code.

fasmat: The comment doesn't address my problem with this.

If I specify that the config file at path `/opt/spacemesh/cfg.ini` shall be used, the program will just use default values and print that the file is missing.

Depending where the `log.Warning` outputs, I might not even see this warning

fasmat: I know according to comments further above this is probably intended which I find even more concerning; the check doesn't even verify that it is actually a "missing file error" just that is not a `flags.IniError`.

config/config.go
poszu: It is intentional. It makes sure that command-line flags are "on top". 

Moved additional ParseFlags out to `poetMain()`.

PR #112, opened 2022-07-28
make poet aware of epoch duration

dshulyak:
related: https://github.com/spacemeshos/pm/issues/143

in order to coordinate poet with epoch duration this change introduces following configuration:
- genesis time (--genesis-time)
- epoch duration (--epoch-duration)
- phase shift (--phase-shift)
- cycle gap (--cycle-gap)
 
poet round execution is started at genesis + epoch_duration * epoch + phase shift
poet round ends at genesis + epoch_duration * (epoch+1) + phase shift - cycle gap
this way poet round is executed for epoch duration - cycle gap

during the cycle gap smeshers that participated in the previous round should prepare a new challenge and register in the new round. smeshers that didn't participate in the previous round may register in the currently open round at any time.

if poet crashes (or terminated) during round - it will recover and finish execution for interrupted round. as long as it recovers before rounds end - smeshers that are using this particular instance of poet will receive a valid poet proof, probably with a lower number of ticks.


dshulyak:
@countvonzero can you please review this change?

service/round.go
```diff
@@ -11,19 +18,16 @@ import (
...
 
 type executionState struct {
-	NumLeaves     uint64
+	Epoch         uint32
 	SecurityParam uint8
 	Members       [][]byte
 	Statement     []byte
 	ParkedNodes   [][]byte
 	NextLeafID    uint64
+	NumLeaves     uint64
```
countvonzero, line 29: is it not possible to derive one from the other (NextLeafID and NumLeaves)?

service/round.go
```diff
@@ -11,19 +18,16 @@ import (
...
 )
 
 type executionState struct {
-	NumLeaves     uint64
+	Epoch         uint32
 	SecurityParam uint8
 	Members       [][]byte
 	Statement     []byte
 	ParkedNodes   [][]byte
 	NextLeafID    uint64
```
countvonzero: can you point me to the code where NextLeafID is updated? i cannot find it

service/service.go
```diff
@@ -212,40 +231,55 @@ func (s *Service) Start(b Broadcaster) error {
...
 			if s.openRound.isEmpty() && !s.cfg.ExecuteEmpty {
 				continue
 			}
 
 			s.prevRound = s.openRound
-			s.openRound = s.newRound()
-			log.Info("Round %v opened", s.openRound.ID)
+			open := time.Now()
+			// TODO(dshulyak) some time is wasted here to persist data on disk
+			// on my computer ~20ms
```
countvonzero, line 272: wasted how? you mean it delays the broadcasting by 20ms? maybe this comment should be on line 281?

config.go
```diff
@@ -95,17 +96,18 @@ func loadConfig() (*config, error) {
 		RawRPCListener:  fmt.Sprintf("localhost:%d", defaultRPCPort),
 		RawRESTListener: fmt.Sprintf("localhost:%d", defaultRESTPort),
 		Service: &service.Config{
-			N:                        defaultN,
```
countvonzero, line 98: since you are removing the config N/InitialRoundDuration, please also update comments that reference these configuration. otherwise this codebase is made more confusing by this PR

service/service.go
```diff
@@ -287,36 +329,30 @@ func (s *Service) Recover() error {
...
+		go func(r *round, rs *roundState) {
 			defer func() {
 				s.Lock()
 				delete(s.executingRounds, r.ID)
 				s.Unlock()
 			}()
 
-			if err = r.recoverExecution(state.Execution); err != nil {
+			end := s.genesis.
+				Add(s.cfg.EpochDuration * time.Duration(r.execution.Epoch+1)).
```
countvonzero: :s/r.execution.Epoch/r.Epoch()/

countvonzero (approved):

service/round.go
dshulyak: it was updated here https://github.com/spacemeshos/poet/blob/a52a2daaa52c43a9a5a204a4bf04a0213c53f878/service/round.go#L222

service/round.go
dshulyak: i avoided that because there was a difference in how they were updated (NumLeaves after the end of the execution and NextLeadID was periodically persisted or if round was interrupted). but after a second look i think there are no downsides for merging it

service/service.go
dshulyak: so each round has a deadline at next_round_start - cycle_gap. ideally we would run poet without wasting time till this deadline. but we are wasting some time because of the inefficiency that i described in this todo

PR #113, opened 2022-07-31
Enable scale code for a merkle proof

evt:
This is needed to enable scale codec for poet proofs as per https://github.com/spacemeshos/go-spacemesh/issues/3271


dshulyak:
you will need to broadcast merkle proof encoded with scale (see what is sent to BroadcastProof), otherwise it won't be decodable on the spacemesh side

evt:
@dshulyak is it correct that I need to replace `xdr` with `scale` codec here - https://github.com/spacemeshos/poet/blob/ef8f28a744fc81cd0a9aedaf64ed4b6dd7dcf73d/service/service.go#L476

do we need to make the same change here aswell then - https://github.com/spacemeshos/poet/blob/ef8f28a744fc81cd0a9aedaf64ed4b6dd7dcf73d/service/util.go#L14

dshulyak:
technically you can leave xdr in persist if it is more complex to change, but ideally we should use one lib for clarity

evt:
> technically you can leave xdr in persist if it is more complex to change, but ideally we should use one lib for clarity

let's do it bit by bit. Done for `serializeProofMsg`. Left for persist as would need codec enabled in github.com/spacemeshos/merkle-tree first

dshulyak (approved):

PR #121, opened 2022-09-09
Gofmt code

fasmat:
- Formatted code with Go 1.18s gofmt
- Update dependencies that are not updated by dependabot


fasmat:
`TestNip` keeps failing for some reason, generating less than 150 leaves when 150 are expected. I cannot reproduce this error locally. On my machine this test passes without problems. Any ideas why the test might fail on github?

hash/hash.go
```diff
@@ -1,6 +1,6 @@
 package hash
 
-import "github.com/spacemeshos/sha256-simd"
+import "github.com/minio/sha256-simd"
```
countvonzero: why this update?

shared/shared.go
```diff
@@ -7,8 +7,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/minio/sha256-simd"
 	"github.com/spacemeshos/merkle-tree"
-	"github.com/spacemeshos/sha256-simd"
```
countvonzero: ditto

hash/hash.go
fasmat: This seems to be the only repo where we are still using our (outdated) fork over the upstream version if the library.

countvonzero: was the test failure caused by minio?

fasmat: Unfortunately not, it seems like golang.org/x/net is the culprit :-( at least if I update the dependency tests start failing. I will still put the migration from `github.com/spacemeshos/sha256-simd` to `github.com/minio/sha256-simd` into its own PR.

countvonzero (approved):

PR #129, opened 2022-10-16
Update CI and fix failing tests

### LINKED ISSUE #131: PoET CI should be executed on Windows and MacOS
### Summary

PoET is used as a dependency for go-spacemesh. Since go-spacemesh is built for Linux, MacOS and Windows, PoET should also be able to be built on all 3 platforms:

### Acceptance criteria

- Update CI to build and run tests on all 3 platforms
   - Add standard set of linters & checkers in CI (staticcheck, golangci-lint, test-fmt, etc.)
   - Report passing & failing tests using JUnit Report
- Target Go 1.19 instead of 1.18
- Fix existing tests that might fail on non-Linux platforms

fasmat:
Closes: https://github.com/spacemeshos/poet/issues/131

Additionally the following improvements are made:

- timeouts increased to hopefully stop flaky tests from failing
- close open files when they are not needed any more (fixes most failing windows tests)
  - requires https://github.com/spacemeshos/merkle-tree/pull/24 to be merged
- Fix some data races in the `signal` package
- Fix windows specific problems with shutdown in harness
- ensure that a test verifies that challenges are handled in the round they are supposed to be handled in; rather than in the same (possibly wrong) round
- Use context in some places where hard-coded timeouts were in place for better controlflow


.github/workflows/ci.yml
```diff
@@ -0,0 +1,103 @@
+name: Go
+
+env:
+  go-version: "1.19"
```
poszu, line 4: :+1:

.golangci.yml
```diff
@@ -0,0 +1,262 @@
+# This file contains all available configuration options
```
poszu, line 1: What do you think about "externalizing" this file to some other repo? It could be fetched from there for CI. We would have the same linting requirements for all our Go repos.

Makefile
```diff
@@ -2,15 +2,66 @@ ORG ?= spacemeshos
...
+	make tidy
+	git diff --exit-code || (git --no-pager diff && git checkout . && exit 1)
+.PHONY: test-tidy
+
+test-fmt:
+	git diff --quiet || (echo "\033[0;31mWorking directory not clean!\033[0m" && git --no-pager diff && exit 1)
+	# We expect `go fmt` not to change anything, the test should fail otherwise
+	go fmt ./...
+	git diff --exit-code || (git --no-pager diff && git checkout . && exit 1)
+.PHONY: test-fmt
```
poszu: I keep mistaking these targets with "tidying tests" and "formatting tests". How about renaming these targets to:
```suggestion
check-tidy:
	# Working directory must be clean, or this test would be destructive
	git diff --quiet || (echo "\033[0;31mWorking directory not clean!\033[0m" && git --no-pager diff && exit 1)
	# We expect `go mod tidy` not to change anything, the test should fail otherwise
	make tidy
	git diff --exit-code || (git --no-pager diff && git checkout . && exit 1)
.PHONY: check-tidy

check-fmt:
	git diff --quiet || (echo "\033[0;31mWorking directory not clean!\033[0m" && git --no-pager diff && exit 1)
	# We expect `go fmt` not to change anything, the test should fail otherwise
	go fmt ./...
	git diff --exit-code || (git --no-pager diff && git checkout . && exit 1)
.PHONY: check-fmt
```

Makefile
```diff
@@ -2,15 +2,66 @@ ORG ?= spacemeshos
...
+	# We expect `go mod tidy` not to change anything, the test should fail otherwise
+	make tidy
+	git diff --exit-code || (git --no-pager diff && git checkout . && exit 1)
+.PHONY: test-tidy
+
+test-fmt:
+	git diff --quiet || (echo "\033[0;31mWorking directory not clean!\033[0m" && git --no-pager diff && exit 1)
+	# We expect `go fmt` not to change anything, the test should fail otherwise
+	go fmt ./...
+	git diff --exit-code || (git --no-pager diff && git checkout . && exit 1)
```
poszu: It should do:
```suggestion
	if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then exit 1; fi
```

integration/harness.go
```diff
@@ -69,23 +68,23 @@ func NewHarness(cfg *ServerConfig) (*Harness, error) {
 // The created process is killed, and the temporary
 // directories are removed.
 func (h *Harness) TearDown(cleanup bool) error {
-	if err := h.server.shutdown(cleanup); err != nil {
-		return err
+	if err := h.conn.Close(); err != nil {
+		return fmt.Errorf("failed to close connection: %w", err)
```
poszu: Is it alright to return early, even in case `conn` failed to close? IMO, it should optimistically continue shutting down other things. Especially, there are still some things to `cleanup`.

integration/harness.go
```diff
@@ -102,14 +101,11 @@ func (h *Harness) RESTListen() string {
 
 // connectClient attempts to establish a gRPC Client connection
 // to the provided target.
-func connectClient(target string) (*grpc.ClientConn, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+func connectClient(ctx context.Context, target string) (*grpc.ClientConn, error) {
 	opts := []grpc.DialOption{
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithBlock(),
```
poszu, line 108: Consider adding a timeout. We cannot tell here if the parent `ctx` has any timeout on it.
```suggestion
		grpc.WithTransportCredentials(insecure.NewCredentials()),
		grpc.WithBlock(),
		grpc.WithTimeout(time.Second * 10),
```

integration/server.go
```diff
@@ -137,7 +137,7 @@ func (s *server) start() error {
 		if err != nil {
 			// Don't propagate 'signal: killed' error,
 			// since it's an expected behavior.
-			if !strings.Contains(err.Error(), "signal: killed") {
+			if !strings.Contains(err.Error(), "signal: killed") && errb.Len() > 0 {
```
poszu: NIT, `errb` is a terrible name. Consider `stderr`
```suggestion
			if !strings.Contains(err.Error(), "signal: killed") && stderr.Len() > 0 {
```

poet_test.go
```diff
@@ -36,7 +38,7 @@ func TestHarness(t *testing.T) {
 	r.NoError(err)
 	cfg.Genesis = time.Now()
 
-	h := newHarness(t, cfg)
+	h := newHarness(t, context.TODO(), cfg)
```
poszu: "Go question". Wouldn't it be proper to use `context.Background()` instead?

poet_test.go
```diff
@@ -117,50 +120,50 @@ func TestHarness_CrashRecovery(t *testing.T) {
...
-			} else {
-				req.Equal(roundsID[roundIndex], res.RoundId)
-			}
+			req.Equalf(roundsID[roundIndex], res.RoundId, "round id mismatch for challenge %d", i)
 		}
 	}
 
-	waitNewRound := func(h *integration.Harness, currentRoundId string) {
-		timeout := time.After(cfg.EpochDuration)
+	waitNewRound := func(h *integration.Harness, ctx context.Context, currentRoundId string) {
```
poszu: Shouldn't the `ctx` be the first argument?

poet_test.go
```diff
@@ -199,10 +202,13 @@ func TestHarness_CrashRecovery(t *testing.T) {
 	req.Equal(roundsID[1], info.ExecutingRoundsIds[0])
 	req.Equal(roundsID[2], info.OpenRoundId)
 	req.NoError(h.TearDown(true))
+
+	// leave some time for the server to shutdown.
+	time.Sleep(100 * time.Millisecond)
```
poszu, line 207: 😢 
It should not spawn the server as a separate process in the first place. But this is a bigger refactor I guess...

service/round_test.go
```diff
@@ -62,64 +56,53 @@ func TestRound_Recovery(t *testing.T) {
...
+	r2recovery1 := newRound(sig, tmpdir, 1)
 	req.Equal(len(challenges), r2recovery1.numChallenges())
 	req.False(r2recovery1.isEmpty())
 
 	state, err := r2recovery1.state()
 	req.NoError(err)
 
 	go func() {
-		time.Sleep(duration / 3)
+		time.Sleep(duration / 5)
```
poszu, line 76: 😢

service/service.go
```diff
@@ -62,7 +61,8 @@ type Service struct {
 
 	// openRound is the round which is currently open for accepting challenges registration from miners.
 	// At any given time there is one single open round.
-	openRound *round
+	openRound      *round
+	openRoundMutex sync.Mutex
```
poszu: How about using an atomic pointer instead?:
```suggestion
        //SAFETY: the contents behind the pointer are read-only
	openRound atomic.Pointer[round]
```

---
Edit: After looking further in the code it seems that this mutex doesn't just guard `openRound` variable only (it is locked for some functions invocations). Please consider documenting what this mutex guards.

service/service.go
```diff
@@ -269,8 +273,10 @@ func (s *Service) Start(b Broadcaster) error {
 			open := time.Now()
 			// TODO(dshulyak) some time is wasted here to persist data on disk
 			// on my computer ~20ms
-			s.openRound = s.newRound(s.prevRound.Epoch() + 1)
-			log.Info("Round %v opened. took %v", s.openRound.ID, time.Since(open))
+			s.openRoundMutex.Lock()
+			s.newRound(s.prevRound.Epoch() + 1)
```
poszu, line 277: Must the whole invocation of `s.newRound` be in a critical section?
If not, consider returning `*round`  and do the assignment under a lock (or use atomic)
If yes, consider documenting that fact in the method's comment.

service/service.go
```diff
@@ -427,15 +435,23 @@ func (s *Service) Info() (*InfoResponse, error) {
 	return res, nil
 }
 
-func (s *Service) newRound(epoch uint32) *round {
+func (s *Service) newRound(epoch uint32) {
```
poszu, line 439: Like that:
```suggestion
// SAFETY: Must be called with s.openRoundMutex locked.
func (s *Service) newRound(epoch uint32) {
```

service/service.go
```diff
@@ -427,15 +435,23 @@ func (s *Service) Info() (*InfoResponse, error) {
...
-	return r
+
+	s.openRound = r
+}
+
+func (s *Service) openRoundID() string {
+	s.openRoundMutex.Lock()
+	defer s.openRoundMutex.Unlock()
+
+	return s.openRound.ID
```
poszu: This method might be called concurrently from many incoming GRPC requests and it just reads. Please consider using an `RWMutex` to allow many concurrent readers. (If you don't go with the atomic.Pointer)

signal/signal.go
```diff
@@ -132,8 +132,12 @@ func (s *Signal) unblockShutdown(id int) {
 	numBlocking := len(s.blocking)
```
poszu: I know it's outside of the scope of this PR, but I don't understand why `s.blocking` is a map instead of just an atomic counter. The shutdown process looks very complicated.

poszu (changes requested):
Great work!

.golangci.yml
fasmat: They differ in the details, i.e. the sorting order for import paths is different for every repository. It might also make sense to disable depreciation warnings in some repos for the whole codebase instead of just for a few files (e.g. the `api` repository).

So while it would be helpful if we could share that file, I'm a bit hesitant since there are some differences.

Makefile
fasmat: I went with consistency since these targets are also named `test-fmt` and `test-tidy` in other repositories. I'm OK with check instead but then we should also update it for other components of spacemesh

Makefile
fasmat: A reasonable change but it if a file isn't properly formatted it doesn't show which file.

poszu: But does it matter? It's a CI check. If it failed, then run `make format` / `gofmt -w .` and done.  It could actually hint to do so:
```suggestion
	if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then echo "Code needs reformatting. Please run 'make format'" && exit 1; fi
```

integration/harness.go
fasmat: This doesn't really do anything: `grpc.WithTimeout` forks the passed in `ctx` by creating a new `context` with the given timeout: https://github.com/grpc/grpc-go/blob/v1.50.1/clientconn.go#L217-L221

EDIT: Just saw you were aware of the timeout being set by context. Should we just assume a default timeout of 10 seconds, i.e. if this wasn't an option of `grpc` would we add the 10 second timeout for the passed in `ctx` ourselves?

EDIT2: `grpc.WithTimeout` is also marked as deprecated with the hint of using `grpc.DialContext` and a context instead.

Makefile
fasmat: For now I leave them as they are for consistency; we can do a rename for all repositories in the future 🙂

integration/server.go
fasmat: `stderr` is already taken by the `StdErrPipe()` of the cmd started. This is a buffer that buffers every byte read from `stderr` (since an `io.Reader` can only be read once). I'll rename it to `stderrBuf`.

poet_test.go
fasmat: `context.TODO()` and `context.Background()` are the same thing. The different names just communicate an intent.

> context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

> context.Background returns a non-nil, empty Context. It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

I was of the opinion that `context.TODO()` would be the appropriate to use in tests, but the documentation says otherwise, then I'll change it.

integration/harness.go
fasmat: It's arguably not an error if the cleanup still works, so instead of returning the error I just log it.

service/service.go
fasmat: Yes, atomic.Pointer wouldn't work without some bigger changes. I changed `openRoundMutex` to be a `sync.RWMutex` so multiple routines can read the ID of the current round.

service/service.go
fasmat: I added clarifying comments.

signal/signal.go
fasmat: Since it turned out not to be a big change I updated the code.

Makefile
```diff
@@ -3,18 +3,16 @@ IMAGE ?= poet
...
 PROTOC_VERSION = 21.8
 PROTOC_GEN_GO_VERSION = v1.28
 PROTOC_GEN_GRPC_VERSION = v1.2
 PROTOC_GEN_GRPC_GATEWAY_VERSION = v1.16.0
 PROTOC_GEN_OPENAPIV2_VERSION = v2.12.0
 
-# `go install` will put binaries in $(GOBIN), avoiding
-# messing up with global environment.
-export GOBIN := $(PWD)/bin
-export PATH := $(GOBIN):$(PATH)
```
fasmat, line 16: I had to remove this or the build fails on non-linux systems: 
https://github.com/spacemeshos/poet/actions/runs/3291366658/jobs/5425450034

countvonzero (commented):
will let Bartosz finish the review. just want to thank you for doing this. 💯

.golangci.yml
poszu: Extra disabled warnings I understand (as temporary until we clean up). But things like import sorting order - that's exactly what I think should be identical in every repo.

Makefile
poszu: I think it would just need a bit of tweaking to work in Windows and should work on Mac, shouldn't it? I can quickly prepare a version for windows for you, what do you think?

Makefile
```diff
@@ -72,6 +126,20 @@ push:
...
+# Lint .proto files
+lint-protos: $(BUF) $(PROTOC)
+	$(BUF) lint
+.PHONY: lint-protos
+
+# Verify if files built from .proto are up to date.
+check: generate
+	@git add -N $(PROTOC_BUILD_DIRS)
+	@git diff --name-only --diff-filter=AM --exit-code $(PROTOC_BUILD_DIRS) \
+	  || { echo "\nPlease rerun 'make generate' and commit changes.\n"; exit 1; }
```
poszu: This part now only checks the proto-generation part. It should probably also verify if `go generate ./...` needs to be run again. The point of this check is to make sure that the dev did not forget to regenerate code.

.golangci.yml
fasmat: The sorting order is:

```golang
import (
	// std library packages
	"log"

	// modules from `go.mod`
	"google.golang.org/grpc"
	"github.com/spacemeshos/merkle-tree"

	// packages from the same module
	"github.com/spacemeshos/poet/server"
	"github.com/spacemeshos/poet/signal"
)
```

So unless we want to mix packages from the same repo into the everything else category - which I'm opposed to - we have to use a different sorting configuration for every repo.

Makefile
fasmat: The job failed on windows and mac - check the link 🙂.

What is the intention about installing the binaries to `./bin`? `go install pkg@version` ignores the `go.mod` file for building the tool and the dependencies are still cached and built in the same directory as the `poet` module. The only difference is that instead of putting the binary into `$(GOPATH)/bin` it's moved to `./bin`?

poszu (changes requested):

Makefile
poszu: The intention is to not install it globally. What if the developer uses different versions for each repository (which actually happens right now)? Because PATH is prepended with ./bin directory, all commands used in this Makefile will use these local binaries.

Makefile
fasmat: It does. It invokes the `generate` target which both calls `go generate` and `$(BUF) generate`

.golangci.yml
poszu: Right! 😭

integration/harness.go
poszu: > EDIT2: `grpc.WithTimeout` is also marked as deprecated with the hint of using `grpc.DialContext` and a context instead.

Oh, too bad :( Then I would create a child `Context` with timeout and pass it to `grpc.DialContext(dialCtx)` to make sure there is a reasonable timeout for the connection.

Makefile
fasmat: The only tool I could find that is not in the same version everywhere is `protoc` which is not installed via `go install`.

Since `make install` is a prerequisite to `make build` and `make test` anyway, why don't we just add it as a dependency to the `make` targets that need it? `go install` is a no-op anyway if instructed to install the same version as already on the system.

integration/harness.go
poszu: I don't understand.

The code just returns if `h.conn.Close()` failed. My point is that it should store the error and try to proceed with shutdown. Finally return any error it saw during Teardown.

fasmat: No it doesn't, please check again I already replaced the `return` with a `log`

Makefile
poszu: How do you know if other people don't use other versions? IMO installing tools globally, and overwriting whatever is already installed, is wrong.

`protoc-gen-go` and `protoc-gen-grpc-gateway` are also of different versions on poet and api repos.

>go install is a no-op anyway if instructed to install the same version as already on the system.

On my system, it takes 13s to rerun `go install` of a single already installed tool.

Makefile
poszu: But it checks for diff in  $(PROTOC_BUILD_DIRS) only.

Makefile
```diff
@@ -18,35 +23,49 @@ PROTOC_BUILD_DIRS := $(PROTOC_GO_BUILD_DIR) $(PROTOC_OPENAPI_BUILD_DIR)
...
+
+# `go install` will put binaries in $(GOBIN), avoiding
+# messing up with global environment.
+export GOBIN := $(BIN_DIR)
+
+install-buf:
+	@mkdir -p $(BIN_DIR)
+	curl -sSL "https://github.com/bufbuild/buf/releases/download/v$(BUF_VERSION)/buf-$(UNAME_OS)-$(UNAME_ARCH)" -o $(BIN_DIR)/buf
+	@chmod +x $(BIN_DIR)/buf
+.PHONY: buf
```
poszu:
```suggestion
.PHONY: install-buf
```

Makefile
```diff
@@ -18,35 +23,49 @@ PROTOC_BUILD_DIRS := $(PROTOC_GO_BUILD_DIR) $(PROTOC_OPENAPI_BUILD_DIR)
...
+install-protoc: protoc-plugins
+	@mkdir -p $(BIN_DIR)
+ifeq ($(OS),Windows_NT)
+	@mkdir -p $(TMP_PROTOC)
+endif
+	curl -sSL https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-${PROTOC_BUILD}.zip -o $(TMP_PROTOC)/protoc.zip
+	@unzip $(TMP_PROTOC)/protoc.zip -d $(TMP_PROTOC)
+	@cp -f $(TMP_PROTOC)/bin/protoc $(BIN_DIR)/protoc
+	@chmod +x $(BIN_DIR)/protoc
+.PHONY: protoc
```
poszu:
```suggestion
.PHONY: install-protoc
```

Makefile
```diff
@@ -56,19 +75,24 @@ protoc-plugins:
...
+all: build
 .PHONY: all
 
 test:
+ifeq ($(UNAME_OS),Darwin)
+	# Workaround for PATH not working on macOS
+	$(BIN_DIR)/gotestsum -- -timeout 5m -p 1 ./...
+else
 	gotestsum -- -timeout 5m -p 1 ./...
+endif
```
poszu: the workaround wouldn't be needed if you just did:
```
GOTESTSUM := $(BIN_DIR)/gotestsum
```

And: 

```suggestion
	$(GOTESTSUM) -- -timeout 5m -p 1 ./...
```

Makefile
```diff
@@ -127,19 +159,14 @@ push:
...
-lint-protos: $(BUF) $(PROTOC)
-	$(BUF) lint
-.PHONY: lint-protos
-
 # Verify if files built from .proto are up to date.
-check: generate
-	@git add -N $(PROTOC_BUILD_DIRS)
-	@git diff --name-only --diff-filter=AM --exit-code $(PROTOC_BUILD_DIRS) \
+test-generate: generate
+	@git add -N .
```
poszu: This line will add **all** _untracked_ files a dev has to _tracked_ files, including any scratch files etc one might have locally. It will mess with ppl's workspace :(

At least leave a warning in a comment...

poszu (approved):
IMO the tools installation steps could be "un-phonied" by making tool targets point to the actual file like it was with `$(BUF)` and `$(protoc)` before.

However, it's not worth the time needed to invest in making it work on all platforms. I'm approving to unblock further work after an offline discussion with @fasmat.

PR #132, opened 2022-10-20
Reproducible protos builds

poszu:
### Motivation
Previously, there were separate `gen_protos.sh` files for building `.proto` files.  They did not contain versions of protoc and its plugins making a reproducible build impossible.

💡prework for https://github.com/spacemeshos/go-spacemesh/issues/3458.

### Changes
Added [buf](https://docs.buf.build/introduction) dependency to ease the process of building, managing dependencies (like `googleapis`), linting etc.

Added Makefile targets to fetch dependencies, build and lint protos and check if files built from protos are up to date.

Added CI job to lint protos and check if files built from protos are up to date.

`protoc`, `buf`, protoc plugins versions are pinned to specific versions to allow reproducible builds.


Makefile
```diff
@@ -24,3 +70,20 @@ docker:
...
+generate: $(BUF) $(PROTOC)
+	$(BUF) generate
+.PHONY: generate
+
+# Verify if files built from .proto are up to date.
+check: generate
+	@git add -N $(PROTOC_BUILD_DIRS)
+	@git diff --name-only --diff-filter=AM --exit-code $(PROTOC_BUILD_DIRS) \
+	  || { echo "\nPlease rerun 'make generate' and commit changes.\n"; exit 1; }
+.PHONY: check
```
poszu:
```suggestion
.PHONY: check

```

buf.gen.yaml
```diff
@@ -0,0 +1,21 @@
+version: v1
+managed:
+  enabled: true
+  go_package_prefix:
+    default: github.com/spacemeshos/poet/gen/proto/go
```
poszu:
```suggestion
    default: github.com/spacemeshos/poet/release/proto/go
```

Makefile
poszu:
```suggestion
.PHONY: check

```

fasmat (approved):

PR #134, opened 2022-10-24
Improvements in Dockerfile

poszu:
- use Go 1.19
- cache downloaded go modules
- use newer alpine (3.16 vs 3.15)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.19-alpine3.16 as build
+FROM golang:1.19-alpine as build
 RUN apk add libc6-compat gcc musl-dev
 WORKDIR /build/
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,6 +1,11 @@
...
 RUN apk add libc6-compat gcc musl-dev
 WORKDIR /build/
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
 COPY . .
 RUN --mount=type=cache,target=/go/pkg/mod go mod download
 RUN --mount=type=cache,target=/root/.cache/go-build go build -o /build/poet
```
fasmat: A second `go mod download` is probably not needed and we have a Makefile target for `build`.

```suggestion
RUN --mount=type=cache,id=build,target=/root/.cache/go-build make build
```

Dockerfile
```diff
@@ -1,6 +1,11 @@
-FROM golang:alpine3.15 as build
+FROM golang:1.19-alpine3.16 as build
```
fasmat: Do we need to specify the alpine version? In `go-spacemesh` we just use:

```suggestion
FROM golang:1.19-alpine as build
```

We only need to make sure Go 1.19 is used.

fasmat (approved):

PR #135, opened 2022-10-27
Extend Submit GRPC API

### LINKED ISSUE #145: extend Poet Submit API

poszu:
Part of https://github.com/spacemeshos/go-spacemesh/issues/3458
Closes #145 

Extended Poet's GRPC API - `Submit()`  method so that challenge details can be passed.

For now, Poet still supports only the old API (passing challenge hash). The signature is verified but no action is performed if it is invalid.

Further verification of the challenge's contents and actually using it will be part of subsequent PRs.


signing/sign.go
```diff
@@ -0,0 +1,96 @@
+package signing
+
+import (
+	"bytes"
+	"crypto"
+	"crypto/ed25519"
```
dshulyak: note that ed25519 that is used for spacemesh protocol signatures is not compliant with canonical ed25519 signature

based on the description in the linked issue:
> Verify signature and extract pubkey (as NodeID).

this change should use modified ed25519 instead of canonical

rpc/api/api.proto
```diff
@@ -69,11 +69,44 @@ message UpdateGatewayResponse {
...
+                uint64 labels_per_unit = 5;
+                uint32 k1 = 6;
+                uint32 k2 = 7;
+            }
+            Proof proof = 1;
+            Metadata metadata = 2;
+        }
+        bytes node_id = 1;
+        bytes positioning_atx_id = 2;
+        bytes pub_layer_id = 3;
```
poszu: It should be uint32

shared/types.go
```diff
@@ -0,0 +1,21 @@
...
+
+type InitialPost struct {
+	Proof    shared.Proof
+	Metadata shared.ProofMetadata
+}
+
+type Challenge struct {
+	NodeID           []byte
+	PositioningAtxId []byte
+	PubLayerId       []byte
```
poszu: It should be uint32.

rpc/api/api.proto
```diff
@@ -69,11 +69,44 @@ message UpdateGatewayResponse {
...
+                uint32 num_units = 3;
+                uint32 bits_per_label = 4;
+                uint64 labels_per_unit = 5;
+                uint32 k1 = 6;
+                uint32 k2 = 7;
+            }
+            Proof proof = 1;
+            Metadata metadata = 2;
+        }
+        bytes node_id = 1;
```
moshababo: `node_id` can be extracted from the signature, no?

rpc/api/api.proto
```diff
@@ -69,11 +69,44 @@ message UpdateGatewayResponse {
...
+        bytes positioning_atx_id = 2;
+        uint32 pub_layer_id = 3;
+        oneof prev {
+            InitialPost initial_post = 4;
+            bytes prev_atx_id = 5;
+        }
+    }
+
+    Data data = 2;
+    bytes pubkey = 3;
```
moshababo: How does `pubkey` differs from `node_id` (within `data`)?

signing/sign.go
```diff
@@ -0,0 +1,96 @@
...
+func (notHashed) HashFunc() crypto.Hash { return crypto.Hash(0) }
+
+type encodable[P any] interface {
+	scale.Encodable
+	*P
+}
+
+// Sign signs data with given signer.
+// *T must implement scale.Encodable which is constrained by Encodable.
+func Sign[T any, Encodable encodable[T]](data T, signer crypto.Signer, pubkey []byte) (Signed[T], error) {
```
moshababo: If `Sign` and `notHashed` were implemented only for testing, they can be removed to the test file.

signing/sign.go
```diff
@@ -0,0 +1,96 @@
...
+
+// NewFromScaleEncodable constructs Signed[T] from a T.
+// *T must implement scale.Encodable which is constrained by Encodable.
+func NewFromScaleEncodable[T any, Encodable encodable[T]](data T, signature, pubkey []byte) (Signed[T], error) {
+	// Serialize it for signature verification
+	var dataBuf bytes.Buffer
+	if _, err := Encodable(&data).EncodeScale(scale.NewEncoder(&dataBuf)); err != nil {
+		return nil, err
+	}
+	if l := len(pubkey); l != ed25519.PublicKeySize {
```
moshababo: A simpler option: 
```go
if len(pubkey) != ed25519.PublicKeySize 
```

rpc/api/api.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/spacemeshos/poet/shared"
+	"github.com/spacemeshos/poet/signing"
+)
+
+var (
+	ErrOneOfNotSet      = errors.New("either initial_post or prev_atx must be set")
+	ErrInvalidPublicKey = errors.New("signer.Public() should return []byte")
+)
+
+func intoSubmitRequestData(d *shared.Challenge) (*rpcapi.SubmitRequest_Data, error) {
```
moshababo, line 18: Same here—if used only in test, I suggest to move it to a test file.

signing/sign.go
```diff
@@ -0,0 +1,96 @@
...
+	return &signedData[T]{
+		data:      data,
+		pubkey:    pubkey,
+		signature: signature,
+	}, nil
+}
+
+// NewFromScaleEncodable constructs Signed[T] from a T.
+// *T must implement scale.Encodable which is constrained by Encodable.
+func NewFromScaleEncodable[T any, Encodable encodable[T]](data T, signature, pubkey []byte) (Signed[T], error) {
```
moshababo: It is recommended to [Accept interfaces, return structs](https://medium.com/@cep21/what-accept-interfaces-return-structs-means-in-go-2fe879e25ee8).

shared/types.go
```diff
@@ -0,0 +1,23 @@
+package shared
```
moshababo, line 1: Can the types here be imported instead of re-defined?

moshababo (commented):
Looks good!
First time I see generics in Go ;)

signing/sign.go
poszu: It's supposed to be used by go-spacemesh for signing the challenge data as well.

signing/sign.go
poszu: 👍 IDK what I had on mind here 😀

signing/sign.go
poszu: Thanks for the link 😉 

I need to accept a concrete instance of `data` (type `T`) here because I want to place it inside of `signedData[T].data`.

poszu: While it is possible to accept `data scale.Encodable` and later do `data.(T)`, the API would lose explicitness (I don't want it to accept **any** encodable data, but rather `T` only. It would also introduce a dynamic type check (panic if `data.(T)` fails) instead of a compile-time type check.

shared/types.go
poszu: I don't understand. Where would they be Imported from? AFAIK `ATXID` and `LayerID` are defined in `go-spacemesh` and Poet should not depend on go-spacemesh. The `Challenge struct` is a Poet-specific type structure.

signing/sign.go
poszu: It was decided in an offline discussion that the poet will use the custom ed25519 implementation.

rpc/api/api.go
poszu: It's part of the public api meant to be used by go-spacemesh.

rpc/api/api.proto
poszu: removed

rpc/api/api.proto
poszu: It's the same. removed after offline discussion.

signing/sign.go
moshababo: Agree, and sorry if it wasn't clear—I was referring to the returning of an interface instead of a concrete type (and not to the acceptance of a concrete type instead of an interface, which many times makes sense).

signing/sign.go
moshababo: Sorry, got confused about the dependency direction between `go-spacemesh` and `poet`.

rpc/api/api.go
moshababo: https://github.com/spacemeshos/poet/pull/135#discussion_r1012818804

shared/types.go
moshababo: https://github.com/spacemeshos/poet/pull/135#discussion_r1012818804

signing/sign.go
poszu: OK, I understand now. The problem is that the returned type is private. I find it wrong to return a private type from a public function.

moshababo: Ok, makes sense then

moshababo (approved):

PR #137, opened 2022-10-28
Refactor Service creation and fix data race on broadcaster access

poszu:
- refactored `service::NewService()` method. It no longer starts service on its own nor creates a `Broadcaster`,
- `service::Service` must be started manually via `service::Service::Start(Broadcaster)`,
- fixed data race between setting and accessing `service::Service::broadcaster`.


moshababo (approved):
LGTM

PR #138, opened 2022-10-31
Use Context instead the signal pkg to control lifetime of Poet Service

poszu:
Poet Service now gets a Context when started. The context is cancelled when the Service is supposed to stop. It also receives an errgroup.Group that can be awaited to wait for a full stop of the Service. The Service spawns all its goroutines in this errgroup.


go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	golang.org/x/crypto v0.1.0
 	golang.org/x/net v0.1.0
+	golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9
```
fasmat:
```suggestion
	golang.org/x/sync v0.1.0
```

prover/prover.go
```diff
@@ -91,7 +88,7 @@ func GenerateProofWithoutPersistency(
 	securityParam uint8,
 	minMemoryLayer uint,
 ) (uint64, *shared.MerkleProof, error) {
-	return GenerateProof(sig, datadir, labelHashFunc, merkleHashFunc, limit, securityParam, minMemoryLayer, persist)
+	return GenerateProof(context.TODO(), datadir, labelHashFunc, merkleHashFunc, limit, securityParam, minMemoryLayer, persist)
```
fasmat: Would it be much effort to pass in a `context` from above here? I think the only place beside tests `GenerateProofWithoutPersistency` is called is in `rpcserver.go` and the method calling it already has a `context`.

rpc/rpcserver.go
```diff
@@ -58,7 +61,9 @@ func (r *rpcServer) Start(ctx context.Context, in *api.StartRequest) (*api.Start
 		return nil, err
 	}
 
-	if err := r.s.Start(b); err != nil {
+	serviceCtx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
```
fasmat: Can we just use the `context` that's passed in instead of the `NotifyContext`?  The caller of `Start` imho should decide when and how it is canceled.

server/server.go
```diff
@@ -55,18 +55,34 @@ func StartServer(cfg *config.Config) error {
...
+		rpcServer := rpccore.NewRPCServer(stop, cfg.DataDir)
 		grpcServer = grpc.NewServer(options...)
 
 		apicore.RegisterPoetCoreProverServer(grpcServer, rpcServer)
 		apicore.RegisterPoetVerifierServer(grpcServer, rpcServer)
 		proxyRegstr = append(proxyRegstr, apicore.RegisterPoetCoreProverHandlerFromEndpoint)
 		proxyRegstr = append(proxyRegstr, apicore.RegisterPoetVerifierHandlerFromEndpoint)
 	} else {
-		svc, err := service.NewService(sig, cfg.Service, cfg.DataDir)
+		svc, err := service.NewService(serverGroup, cfg.Service, cfg.DataDir)
```
fasmat: imho `NewService` should not receive the errGroup as parameter, but rather just the context and use it to create its own `errGroup`. goroutines started by different components share the same `errGroup`, which could lead to problems by components interfering with each other.

service/round.go
```diff
@@ -65,15 +65,14 @@ type round struct {
...
-	sig       *signal.Signal
 	submitMtx sync.Mutex
 }
 
 func (r *round) Epoch() uint32 {
 	return r.execution.Epoch
 }
 
-func newRound(sig *signal.Signal, datadir string, epoch uint32) *round {
+func newRound(ctx context.Context, serverGroup *errgroup.Group, datadir string, epoch uint32) *round {
```
fasmat: same here, I think a round should have its own `errgroup` instead of having it passed in from the parent.

service/service.go
```diff
@@ -133,103 +129,55 @@ func NewService(sig *signal.Signal, cfg *Config, datadir string) (*Service, erro
...
 				return nil, err
 			}
 		}
 	}
 
-	state, err := s.state()
+	state, err := state(datadir)
 	if err != nil {
-		if !strings.Contains(err.Error(), "file is missing") {
+		if !errors.Is(err, ErrFileIsMissing) {
```
fasmat: ❤️

fasmat (commented):
Good work! I added some small suggestions to make the coupling between components a bit looser.

prover/prover.go
poszu: It shouldn't be much work, I can do it. I used a `TODO()` Context here to mimick previous behaviour (previously it used a dummy global `sig` which never shut down).

rpc/rpcserver.go
poszu: No, that's not possible :) The context, it's called in, is a GRPC request. It will be cancelled when the GRPC call finishes. That's not what is needed. Here, I want to create a long-living context for the lifetime of the Service, which is until the application closes (the Service cannot be stopped).

server/server.go
poszu: The reason to share this error group is to provide a graceful shutdown, which would allow all goroutines to finish. It's required to let a certain state to be persisted when the application closes (when the context is cancelled). 

Previously it was achieved via calls to `BlockShutdown`. Now, this synchronization is done via calling `.Wait()` on the errgroup.

rpc/rpcserver.go
fasmat: Then this feels out of place here. Maybe the `NotifyContext` can then be created within `r.s.Start`?

server/server.go
fasmat: I am aware 🙂, but this way if a routine by `service` errors the `server` also shuts down. Previously this would only happen on a shutdown signal.

Instead I think errors should bubble up and the parent should decide if they want to restart the service or also shutdown.

prover/prover.go
poszu: Done

rpc/rpcserver.go
poszu: Refactored, now SIGINT is handled at a single place in server/server.go. The entity spawning a Service is directly responsible for calling `Service::Shutdown()`. I followed the design of `http.Server` from the stdlib here.

server/server.go
poszu: Refactored. `Service` now holds its own `errgroup` inside, similar to `http.Server` from the stdlib.

service/round_test.go
```diff
@@ -209,10 +210,12 @@ func TestRound_State(t *testing.T) {
...
 	r.broadcasted()
-	time.Sleep(1 * time.Second)
+	req.NoError(r.waitTeardown(context.Background()))
 
 	// Verify cleanup.
 	state, err = r.state()
 	req.EqualError(err, fmt.Sprintf("file is missing: %v", filepath.Join(r.datadir, roundStateFileBaseName)))
 	req.Nil(state)
+
+	stop()
```
fasmat, line 220: When creating `context.WithCancel` you can just call `defer stop()` to make sure it is called at least once (multiple calls don't cause problems). Otherwise this might not be called when a `require` check fails.

service/round_test.go
```diff
@@ -74,35 +74,35 @@ func TestRound_Recovery(t *testing.T) {
...
 	state, err = r2recovery2.state()
 	req.NoError(err)
 
-	req.NoError(r2recovery2.recoverExecution(state.Execution, time.Now().Add(duration)))
+	req.NoError(r2recovery2.recoverExecution(ctx, state.Execution, time.Now().Add(duration)))
 
 	// Request shutdown.
-	sig.RequestShutdown()
-	time.Sleep(100 * time.Millisecond)
+	stop()
```
fasmat, line 94: same as with the other test

service/service_test.go
```diff
@@ -123,21 +122,24 @@ func TestService_Recovery(t *testing.T) {
...
-	s, err = NewService(sig, cfg, tempdir)
+	s, err = NewService(cfg, tempdir)
 	req.NoError(err)
 
 	err = s.Start(broadcaster)
 	req.NoError(err)
 
 	// Service instance should recover 2 rounds: round 1 in executing state, and round 2 in open state.
-	req.Equal(1, len(s.executingRounds))
+	req.Eventually(func() bool { s.Lock(); defer s.Unlock(); return len(s.executingRounds) == 1 }, time.Second, time.Millisecond)
```
fasmat, line 137: Does this fix one of our flaky tests? If yes good find 👍 

I would just increase the check interval from `time.Millisecond` to maybe `20 * time.Millisecond` to give more time to the service where the mutex isn't locked.

service/service_test.go
```diff
@@ -123,21 +122,24 @@ func TestService_Recovery(t *testing.T) {
...
+	s.Lock()
 	first, ok := s.executingRounds["0"]
+	s.Unlock()
 	req.True(ok)
-	req.Equal(s.openRoundID(), "1")
+	req.NotNil(s.openRoundID())
+	req.Equal(*s.openRoundID(), "1")
```
fasmat: This calls `s.openRound()` twice. What's the reason for returning a pointer instead o the value?

If it we stick with the pointer this should avoid a panic and calling `openRoundID()` twice.

```suggestion
	openRoundID := s.openRoundID()
	req.NotNil(openRoundID)
	req.Equal("1", *openRoundID)
```

Also the first parameter is `expected` and the second `actual` 😉. Otherwise this might lead to confusing messages when the assert fails.

service/service.go
```diff
@@ -388,29 +423,29 @@ func (s *Service) Info() (*InfoResponse, error) {
...
 	if err := saveState(s.datadir, s.privKey); err != nil {
 		panic(err)
 	}
-	r := newRound(s.sig, s.datadir, epoch)
+	r := newRound(ctx, s.datadir, epoch)
 	if err := r.open(); err != nil {
 		panic(fmt.Errorf("failed to open round: %v", err))
 	}
 
 	s.openRound = r
```
fasmat: assignment without mutex protection or is it not required here?

service/service.go
```diff
@@ -301,14 +333,15 @@ func (s *Service) Recover() error {
 				Add(s.cfg.PhaseShift).
 				Add(-s.cfg.CycleGap)
 
-			if err = r.recoverExecution(rs.Execution, end); err != nil {
+			if err = r.recoverExecution(ctx, rs.Execution, end); err != nil {
 				s.asyncError(fmt.Errorf("recovery: round %v execution failure: %v", r.ID, err))
-				return
+				return err
```
fasmat: Is `asyncError` now even still required? If not we could just change that to:

```suggestion
			if err := r.recoverExecution(ctx, rs.Execution, end); err != nil {
				err = fmt.Errorf("recovery: round %v execution failure: %w", r.ID, err)
				log.Error(err.Error())
				return err
			}
```

Is `s.errChan` still in use?

service/service.go
```diff
@@ -170,16 +169,75 @@ func NewService(sig *signal.Signal, cfg *Config, datadir string) (*Service, erro
...
+	s.Lock()
+	if !s.started.CompareAndSwap(false, true) {
+		s.Unlock()
 		return ErrAlreadyStarted
 	}
 
+	// Create a context for the running Service.
+	// This context will be canceled when Service is stopped.
+	ctx, stop := context.WithCancel(context.Background())
+	runningGroup, _ := errgroup.WithContext(ctx)
```
fasmat: Hm... I'm not sure about this. The idea of the `context` returned by `errgroup.WithContext` is that it will be canceled by the `errgroup` automatically if one of the go routines insides returns an error. So the context returned by the `errgroup` imho should be passed into `runningGroup.Go(func() {})` over any other context, since otherwise the other routines don't stop as early as possible.

We can still use the parent context to cancel explicitly, but the routines started with `runningGroup.Go` should listen to the context returned by `errgroup.WithContext`.

service/service.go
```diff
@@ -170,16 +169,75 @@ func NewService(sig *signal.Signal, cfg *Config, datadir string) (*Service, erro
 	return s, nil
 }
 
+func (s *Service) loop(ctx context.Context) {
+	executingRounds, _ := errgroup.WithContext(ctx)
```
fasmat: See my other comment about the context returned from `errgroup.WithContext`.

service/service.go
```diff
@@ -170,16 +169,75 @@ func NewService(sig *signal.Signal, cfg *Config, datadir string) (*Service, erro
...
+		prevRound := s.openRound
+		open := time.Now()
+		s.newRound(ctx, prevRound.Epoch()+1)
+		s.openRoundMutex.Unlock()
+		log.Info("Round %v opened. took %v", s.openRoundID(), time.Since(open))
+
+		executingRounds.Go(func() error {
+			round := prevRound
+			if err := s.executeRound(ctx, round); err != nil {
+				s.asyncError(fmt.Errorf("round %v execution error: %v", round.ID, err))
```
fasmat, line 216: see my other comment about `asyncError` and `errChan`.

service/service.go
```diff
@@ -170,16 +169,75 @@ func NewService(sig *signal.Signal, cfg *Config, datadir string) (*Service, erro
...
+				return nil
```
fasmat, line 217: I assume this should return an error instead of `nil`?

service/service.go
```diff
@@ -52,11 +52,14 @@ type serviceState struct {
...
 type Service struct {
+	runningGroup *errgroup.Group
+	stop         context.CancelFunc
+
 	cfg            *Config
 	datadir        string
 	genesis        time.Time
 	minMemoryLayer uint
-	started        int32
+	started        atomic.Bool
```
fasmat, line 65: 👍

fasmat (changes requested):
I added a few comments specifically surrounding the context from `errGroup.WithContext` and it's usage over the parent context for spawning go routines.

service/service_test.go
poszu: It's better but still can be flaky because the round expected to be executing might already finish. A millisecond is still a long time, increasing it will make it more flaky :(

IMO the tests generally shouldn't inspect the inner (private) state of Service which is extremely asynchronous (rounds start and finish asynchronously) but rather make assertions at the observable API level. But that's a bigger tests refactor.

service/service_test.go
poszu: Calling `openRoundID()` twice is not that big of a deal, it's just a getter.

I made it return a pointer because `s.openRound` **is not always set**. It's set after `Service::Start()` returns. It's a problem in the async API. An RPC call `GetInfo()` could panic because of it (a segfault).

Anyway - I rewritten `Start()` to set `Service::started bool` only after the Service is completely started (and `Service::openround` is set) and reverted to returning a `string` here, which is safe now.

service/round_test.go
poszu: I know, but in this test the context is recreated many times and `stop()` is called many times manually. I call the last one manually too for consistency.

service/service.go
poszu: This method assumes to be called under a lock. 
https://github.com/spacemeshos/poet/blob/0a8257879fd821e66791e50b46c3e47edaca5b78/service/service.go#L424-L425

service/service.go
poszu: Unfortunately, tests depend on it.

service/service.go
poszu: As discussed offline - I specifically don't want the auto-cancellation feature. As suggested, I refactored to use a zero-value error group without context.

The spawned goroutines check the parent context (which describes the service running lifetime) for cancellation.

service/service.go
poszu: No, I specifically don't want a failed round to cancel the context and close the Service. An error is logged and Poet Service continues.

fasmat (approved):

service/service.go
poszu: Created an issue https://github.com/spacemeshos/poet/issues/140 for `errChan` not being read anywhere, which will lead to hang goroutines and resources leak.

PR #139, opened 2022-11-02
Challenge validation

### LINKED ISSUE #146: validate submitted challenge

poszu:
Part of https://github.com/spacemeshos/go-spacemesh/issues/3458.
Closes #146 

Added validation of submitted poet challenge according to the rules described in https://github.com/spacemeshos/go-spacemesh/issues/3458.

The new Submit API is still halfway. Missing functionality is mentioned in TODOs. It will come in subsequent PRs.


rpc/rpcserver.go
```diff
@@ -47,19 +55,38 @@ func (r *rpcServer) Start(ctx context.Context, in *rpcapi.StartRequest) (*rpcapi
 		broadcastAcks = 1
 	}
 
+	gtwManager, err := gateway.NewGatewayManager(ctx, in.GatewayAddresses)
+	//lint:ignore SA5001 (need to Close() even in case of an error)
+	defer gtwManager.Close() // nolint:staticcheck // SA5001 (need to Close() even in case of an error)
+	if len(gtwManager.Connections()) < connAcks {
+		return nil, err
+	}
```
fasmat: I am confused by this. If there is at least `connAck` connection the `err` returned by `NewGatewayManager` is ignored? Can this logic be added to `NewGatewayManager` then to only return a non-`nil` `err` when not enough connections were established?

This would also get rid of the need to disable linting and static checking.

rpc/rpcserver.go
```diff
@@ -83,10 +110,19 @@ func (r *rpcServer) UpdateGateway(ctx context.Context, in *rpcapi.UpdateGatewayR
 		broadcastAcks = 1
 	}
 
+	gtwManager, err := gateway.NewGatewayManager(ctx, in.GatewayAddresses)
+	//lint:ignore SA5001 (need to Close() even in case of an error)
+	defer gtwManager.Close() // nolint:staticcheck
+	if len(gtwManager.Connections()) < connAcks {
+		return nil, err
+	}
```
fasmat: See my previous comment.

rpc/rpcserver.go
```diff
@@ -95,23 +131,43 @@ func (r *rpcServer) UpdateGateway(ctx context.Context, in *rpcapi.UpdateGatewayR
...
 
 	return &rpcapi.UpdateGatewayResponse{}, nil
 }
 
 func (r *rpcServer) Submit(ctx context.Context, in *rpcapi.SubmitRequest) (*rpcapi.SubmitResponse, error) {
-	// TODO(brozansk) return an error once migration to new API is complete
-	api.FromSubmitRequest(in)
+	// Temporarily support both the old and new challenge submission API.
+	// TODO(brozansk) remove support for data []byte after go-spacemesh is updated to
+	// use the new API.
```
fasmat, line 149: Is there an issue number that can be referenced here?

service/db.go
```diff
@@ -31,6 +31,10 @@ func (db *LevelDB) Put(key, value []byte) error {
 	return db.DB.Put(key, value, db.wo)
 }
 
+func (db *LevelDB) Has(key []byte) (bool, error) {
+	return db.DB.Has(key, db.ro)
+}
```
fasmat, line 36: Initially I thought that the methods defined here aren't necessary because `struct` embedding should expose them anyway.

This is only true for the `Close()` method though it can be removed from `LevelDB`.

service/service.go
```diff
@@ -389,20 +409,44 @@ func (s *Service) executeRound(ctx context.Context, r *round) error {
...
 	}
 
-	s.openRoundMutex.Lock()
-	r := s.openRound
-	err := r.submit(data)
-	s.openRoundMutex.Unlock()
-	if err != nil {
-		return nil, err
-	}
+	// TODO(brozansk) Remove support for `challenge []byte` eventually
```
fasmat: Is this part of this PR or of another issue?

service/service.go
```diff
@@ -389,20 +409,44 @@ func (s *Service) executeRound(ctx context.Context, r *round) error {
...
+	if signedChallenge != nil {
+		log.Debug("Using the new challenge submission API")
+		// SAFETY: it will never panic as `s.atxProvider` is set in Start
+		atxProvider := s.atxProvider.Load().(types.AtxProvider)
+		if err := validateChallenge(ctx, signedChallenge, atxProvider, s.postConfig.Load(), verifying.Verify); err != nil {
+			return nil, nil, err
+		}
+		// TODO(brozansk) calculate sequence number
+		// TODO(brozansk) calculate hash
```
fasmat: Part of this PR or another issue?

service/service.go
```diff
@@ -389,20 +409,44 @@ func (s *Service) executeRound(ctx context.Context, r *round) error {
...
 
-	return r, nil
+		s.openRoundMutex.Lock()
+		r := s.openRound
+		err := r.submit(signedChallenge.PubKey(), challenge)
+		s.openRoundMutex.Unlock()
+		if err != nil {
+			return nil, nil, err
+		}
+		return r, []byte("TODO(brozansk) calculate hash"), nil
```
fasmat, line 446: Part of this PR or another issue?

gateway/gateway_manager.go
```diff
@@ -0,0 +1,129 @@
...
+	errors := make([]error, 0)
+
+	errorsChan := make(chan error, len(gateways))
+	connsChan := make(chan *grpc.ClientConn, len(gateways))
+
+	opts := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithBlock(),
+		// XXX: this is done to prevent routers from cleaning up our connections (e.g aws load balances..)
+		// TODO: these parameters work for now but we might need to revisit or add them as configuration
```
fasmat, line 93: I know you only copied this, but can we find out where this TODO comes from and create an issue for it? Having these configurable seems important.

go.mod
```diff
@@ -3,16 +3,18 @@ module github.com/spacemeshos/poet
...
+	github.com/hashicorp/golang-lru v0.5.4
 	github.com/jessevdk/go-flags v1.5.0
 	github.com/minio/sha256-simd v1.0.0
 	github.com/nullstyle/go-xdr v0.0.0-20180726165426-f4c839f75077
 	github.com/spacemeshos/api/release/go v1.5.1
 	github.com/spacemeshos/ed25519 v0.0.0-20221025075941-ae6f3a8a6218
 	github.com/spacemeshos/go-scale v1.1.0
 	github.com/spacemeshos/merkle-tree v0.1.0
-	github.com/spacemeshos/post v0.0.0-20221027120329-e896aa763a95
+	github.com/spacemeshos/post v0.0.0-20221028220902-534f3f813233
```
fasmat: Post is already at version v0.2.0

Makefile
```diff
@@ -52,6 +52,21 @@ endif
 export GOBIN := $(BIN_DIR)
 GOTESTSUM := $(GOBIN)/gotestsum
 
+BUILD_DEPS := $(BIN_DIR)/libgpu-setup.so
+export CGO_LDFLAGS := -L$(BIN_DIR)  -Wl,-rpath,$(BIN_DIR)
+
+# We depend on post submodule Makefile.Inc fetching
+# libgpu-setup.so so there is no must
+# to keep it in sync with Go's dependency on post package.
+post/Makefile.Inc:
```
fasmat: This unfortunately doesn't ensure that the submodule is checked out at the same version as the version referenced in `go.mod`.

Makefile
```diff
@@ -52,6 +52,21 @@ endif
...
+	git submodule update --init -- post
+
+$(BIN_DIR)/libgpu-setup.so: post/Makefile.Inc
+	BIN_DIR=$(abspath $(dir $@))/ make -f post/Makefile.Inc get-gpu-setup
```
fasmat: On non-linux platforms there is no `libgpu-setup.so`. For Windows or Mac and calling `make get-gpu-setup` will not create such a file. A file that needs to be present in the `build` folder and is available on all supported platforms is `api.h`.

Calling `get-gpu-setup` will download a `.zip` file to the working directory. Should we add this to the `.gitignore` file?

types/types.go
```diff
@@ -0,0 +1,27 @@
+package types
+
+import (
+	"context"
+
+	"github.com/spacemeshos/poet/shared"
+)
+
+type ATX struct {
```
fasmat, line 9: Probably to big of an issue for this PR: but we should really separate types used within a code base from types used for (de-)serialization of data sent over the wire or persisted in a DB/file.

The later should all be specified by some schema in a separate repository (maybe github.com/spacemeshos/api ?) to be used by all components. They don't have to be in the form of *.proto files, but should still be checked for non-backwards compatible changes when they are updated.

gateway/gateway_manager_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConnecting(t *testing.T) {
+	gtw := NewMockGrpcServer(t)
+	go func() {
+		require.NoError(t, gtw.Serve())
+	}()
+	t.Cleanup(gtw.Stop)
```
fasmat: This is racy. `require.NoError` is evaluated after the test ends. Maybe something like this ensures that the assertion is guaranteed to be part of the test?

```suggestion
	var eg errgroup.Group
	eg.Go(func() error {
		assert.NoError(t, gtw.Serve())
		return nil
	})
	t.Cleanup(func() {
		gtw.Stop()
		eg.Wait()
	})
```

gateway/mock_gateway.go
```diff
@@ -0,0 +1,51 @@
...
+	"google.golang.org/grpc"
+)
+
+// MockGrpcServer allows for simulating a GRPC server in tests.
+// Typical usage:
+// ```
+// gtw := NewMockGrpcServer(t)
+// <register required GRPC services>
+// go func() { require.NoError(t, gtw.Serve()) }()
+// t.Cleanup(gtw.Stop)
```
fasmat: As noted above, this will not check if `gtw.Serve()` returns an error or not. Use a waitgroup, you can even hide it in the implementation of `MockGrpcServer`:

```golang
func (s *MockGrpcServer) Serve() {
	s.eg.Go(func() error { s.Server.Serve(s.listener) })
}

func (s *MockGrpcServer) Stop() error {
    s.Server.Stop()
    return s.eg.Wait()
}
```

And then just assert like this:

```golang
gtw := NewMockGrpcServer(t)
// register required services
gtw.Serve()
t.Cleanup(func() { assert.NoError(gtw.Stop()) })
// do the test
```

service/round.go
```diff
@@ -332,12 +337,12 @@ func (r *round) calcMembersAndStatement() ([][]byte, []byte, error) {
 	iter := r.challengesDb.Iterator()
 	defer iter.Release()
 	for iter.Next() {
-		key := iter.Key()
-		keyCopy := make([]byte, len(key))
-		copy(keyCopy, key)
+		challenge := iter.Value()
+		challengeCopy := make([]byte, len(challenge))
+		copy(challengeCopy, challenge)
```
fasmat, line 342: Why does the challenge need to be copied here?

service/round.go
```diff
@@ -136,14 +136,19 @@ func (r *round) isOpen() bool {
 	return !r.opened.IsZero() && r.executionStarted.IsZero()
 }
 
-func (r *round) submit(challenge []byte) error {
+func (r *round) submit(key, challenge []byte) error {
```
fasmat, line 139: key is sometimes the public key used to sign the challenge and sometimes the challenge itself. 
If I understood the code correctly this is only temporary and will in the future be replaced by just the public key?

service/service.go
```diff
@@ -361,6 +373,14 @@ func (s *Service) SetBroadcaster(b Broadcaster) {
 	}
 }
 
+func (s *Service) SetAtxProvider(provider types.AtxProvider) {
+	s.atxProvider.Store(provider)
+}
+
+func (s *Service) SetPostConfig(config *types.PostConfig) {
+	s.postConfig.Store(config)
+}
```
fasmat, line 382: Can this cause issues where some of the fields of Broadcaster, Provider and PostConfig are updated and not all of them? What if a request is served right between changing the AtxProvider and changing the PostConfig?

imo `SetBroadcaster`, `SetAtxProvider` and `SetPostConfig` should be a single method that sets them atomically.

service/service.go
```diff
@@ -3,6 +3,7 @@ package service
 import (
 	"bytes"
 	"context"
+	"crypto/ed25519"
```
fasmat, line 6: What is the reason of using `crypto/ed25519` over `github.com/spacemeshos/ed25519`?

`golang.or/x/crypto/ed25519` which since Go 1.13 is now just an alias for `crypto/ed25519` is still used in `poet` in some places.

service/service_test.go
```diff
@@ -40,11 +44,14 @@ func TestService_Recovery(t *testing.T) {
 		CycleGap:      time.Second / 4,
 	}
 
+	ctrl := gomock.NewController(t)
+	atxProvider := mock_types.NewMockAtxProvider(ctrl)
```
fasmat: `_` in the package name is unusual for go (see also here: https://go.dev/blog/package-names#package-names). Would it be ok to just rename the `mock_types` package to `mocks`?

fasmat (changes requested):

service/validation_test.go
```diff
@@ -0,0 +1,248 @@
...
+	for _, o := range overrides {
+		o(&challenge)
+	}
+	return &signedChallengeMock{
+		challenge: challenge,
+		nodeID:    []byte{},
+	}
+}
+
+type signedChallengeMock struct {
```
fasmat, line 46: Any particular reason why a handwritten mock over a generated one?

service/service_test.go
```diff
@@ -263,8 +276,9 @@ func TestNewService(t *testing.T) {
 
 	// Submit challenges.
 	for i := 0; i < len(challenges); i++ {
-		round, err := s.Submit(challenges[i].data)
+		round, hash, err := s.Submit(context.Background(), challenges[i].data, nil)
 		req.NoError(err)
+		req.Equal(hash, challenges[i].data)
```
fasmat, line 281: This seems odd to me. Why return an input back as output?

service/service_test.go
```diff
@@ -67,9 +74,10 @@ func TestService_Recovery(t *testing.T) {
 	submittedChallenges := make(map[int][]challenge)
 	submitChallenges := func(roundIndex int, groupIndex int) {
 		challengesGroup := challengeGroups[groupIndex]
-		for i := 0; i < len(challengeGroups[groupIndex]); i++ {
-			round, err := s.Submit(challengesGroup[i].data)
+		for _, challenge := range challengeGroups[groupIndex] {
+			round, hash, err := s.Submit(context.Background(), challenge.data, nil)
 			req.NoError(err)
+			req.Equal(hash, challenge.data)
```
fasmat, line 80: see my previous comment

rpc/rpcserver.go
poszu: Yes - there is a ticket for it #147. I will add the link in the comment.

server/server.go
```diff
@@ -68,26 +71,37 @@ func StartServer(ctx context.Context, cfg *config.Config) error {
 		if err != nil {
 			return err
 		}
-		if len(cfg.Service.GatewayAddresses) > 0 || cfg.Service.DisableBroadcast {
+		gtwManager, _ := gateway.NewGatewayManager(ctx, cfg.Service.GatewayAddresses)
+		if len(gtwManager.Connections()) > 0 {
```
fasmat: If the check for the required number of connections already happens in `NewGatewayManager` this could simplify to:

```golang
gtwManager, err := gateway.NewGatewayManager(ctx, ctg.Service.GateWayAddress, 1)
switch {
case err == nil:
	// no error and enough connections
case errors.Is(err, ErrNotEnoughConnections):
	log.Info("Service not starting, waiting for start request")
default:
	// some other error occurred that might need handling.
}
```

service/service.go
poszu: It's part of #147

gateway/gateway_manager.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (e *ConnectingErrors) Is(target error) bool {
+	_, ok := target.(*ConnectingErrors)
+	return ok
+}
+
+// GatewayManager aggregates GRPC connections to gateways.
+// Its Close() must be called when the connections are no longer needed.
+type GatewayManager struct {
```
fasmat: `gateway.GatewayManager` could be shortened to `gateway.Manager`.

gateway/gateway_manager.go
```diff
@@ -0,0 +1,129 @@
...
+		return
+	}
+	for _, conn := range m.connections {
+		conn.Close()
+	}
+}
+
+// NewGatewayManager creates a GatewayManager connected to `gateways`.
+// Close() must be called when the connections are no longer needed.
+func NewGatewayManager(ctx context.Context, gateways []string) (*GatewayManager, error) {
```
fasmat: `NewManager`?

service/service.go
poszu: other

service/service.go
poszu: other

rpc/rpcserver.go
```diff
@@ -23,10 +29,12 @@ type rpcServer struct {
 var _ rpcapi.PoetServer = (*rpcServer)(nil)
 
 // NewRPCServer creates and returns a new instance of the rpcServer.
-func NewRPCServer(service *service.Service) *rpcServer {
-	return &rpcServer{
-		s: service,
+func NewRPCServer(service *service.Service, gtwManager *gateway.GatewayManager) *rpcServer {
```
fasmat: `NewServer`

Makefile
poszu: Yes, hence the comment.

fasmat (changes requested):

Makefile
fasmat: Why not copy the Makefile instead of having the whole repository as submodule?

service/validation_test.go
poszu: Yes. Gomock doesn't fully support generic interfaces yet.

gateway/mock_gateway.go
poszu: Updated

service/round.go
poszu: Yes. It's temporary to make Poet support both interfaces until go-spacemesh moves to the new API. The key will be the NodeID (pubkey).

service/service.go
poszu: I don't think it's an issue. 

However I just realized of a different one that can actually cause problems. The grpc connections of the old broadcaster and atxProvider can be closed while they are still being used:

1. Thread A: `oldB := s.broadcaster.Load()`
2. Thread B: `s.SetBroadcaster(newB)`
3. Thread B: close GRPC connections for oldB 
4. Thread A: Try to broadcast using `oldB` - no UB but broadcast will fail.

service/service.go
poszu: golang's implementation is used for poet's keys. I just changed the import path here. Our custom implementation is used to verify signature of submitted challenge.

service/service_test.go
poszu: Should be possible

service/service_test.go
poszu: It's a side effect of temporary support of both old and new APIs. It will change after I implement challenge hashing (#148 )

service/round.go
poszu: I don't know _for sure_, but I think it's because the slice points to some _snapshot_ of DB (accessed by the Iterator) and it might be reused/changed after the Iterator is released.

Makefile
poszu: Why make a copy? By having a submodule we can easily keep it up to date and protect these makefiles from diverging (a submodule makes it read-only).

service/service.go
fasmat: OK, but what is the reason not to use `github.com/spacemeshos/ed25519` here? No signing or verification is performed with the `crypto/ed25519` keys.

Using two different implementations of the same algorithm might lead to problems/confusion in the future.

service/round.go
fasmat: In the documentation for `iter.Value()` it says:

> The caller should not modify the contents of the returned slice, and its contents may change on the next call to any 'seeks method'.

So you are probably right.

Makefile
fasmat: If we just need one file from the repository we could also just `wget` this one file:

```bash
Makefile.Inc:
	wget https://raw.githubusercontent.com/spacemeshos/post/v0.2.0/Makefile.Inc
```

This is faster than cloning a whole repository, and we can track the version of post poet depends on within the makefile instead of having to update `.gitmodules`.

service/service.go
fasmat: This generally appears to be dangerous to me to swap out configuration/dependencies while the server is potentially serving requests.

How much effort would it be to start up a second service with the new dependencies and starting to use this instead of the old service and then shutting down the old service when all requests made to it are done or timed out?

service/validation_test.go
fasmat: Do we need the generic type? 😉  The type it is used for appears to always be the same.

service/service.go
```diff
@@ -412,21 +412,29 @@ func (s *Service) executeRound(ctx context.Context, r *round) error {
...
 		}
-		// TODO(brozansk) calculate sequence number
-		// TODO(brozansk) calculate hash
+
+		var sequence uint64
+		if id := signedChallenge.Data().PreviousATXId; id != nil {
+			atx, _ := atxProvider.Get(ctx, id) // Won't fail as challenge is already validated
+			sequence = atx.Sequence + 1
+		}
+		log.Debug("Calculated sequence: %d", sequence)
```
fasmat, line 435: `sequence` is only used to log?

service/service.go
```diff
@@ -412,21 +412,29 @@ func (s *Service) executeRound(ctx context.Context, r *round) error {
...
 		return nil, nil, ErrNotStarted
 	}
 
-	// TODO(brozansk) Remove support for `challenge []byte` eventually
+	// TODO(brozansk) Remove support for `challenge []byte` eventually (https://github.com/spacemeshos/poet/issues/147)
 	if signedChallenge != nil {
 		log.Debug("Using the new challenge submission API")
 		// SAFETY: it will never panic as `s.atxProvider` is set in Start
 		atxProvider := s.atxProvider.Load().(types.AtxProvider)
 		if err := validateChallenge(ctx, signedChallenge, atxProvider, s.postConfig.Load(), verifying.Verify); err != nil {
```
fasmat: since `atxProvider` and `postConfig` can be changed independently; one or both might change between `atxProvider.Load()` and `postConfig.Load()`.

I think these two (along possibly with `boradcaster` should be changed atomically or rare issues might occur where the wrong postConfig is used to validate a challenge from a different atxProvider.

gateway/activation/errors.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (e *AtxNotFoundError) Error() string {
+	return fmt.Sprintf("ATX %X not found", e.id)
+}
+
+func (e *AtxNotFoundError) Is(target error) bool {
+	_, ok := target.(*AtxNotFoundError)
+	return ok
+}
```
fasmat: Imo this should check if id is the same:

```suggestion
func (e *AtxNotFoundError) Is(target error) bool {
	other, ok := target.(*AtxNotFoundError)
	return ok && other.id == e.id
}
```

If one just wants to check if any atx wasn't found this can still be accomplished `errors.As` or `require.ErrorAs`:

```golang
var target AtxNotFoundError
require.ErrorsAs(err, &target) // an AtxNotFoundError with any ID
```

gateway/activation/errors.go
```diff
@@ -0,0 +1,46 @@
...
+	source error
+}
+
+func (e *TransportError) Error() string {
+	return fmt.Sprintf("transport error: %s (%v)", e.msg, e.source)
+}
+
+func (e *TransportError) Unwrap() error { return e.source }
+
+func (e *TransportError) Is(target error) bool {
```
fasmat: Same as above, this checks if one `TransportError` is the same as any other `TransportError`, not if they are actually equal, for this `msg` and `source` should be checked for equality.

If just the type is relevant for the check, `errors.As` is imo the better check.

gateway/activation/atx_provider.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (a *roundRobinAtxProvider) Get(ctx context.Context, id shared.ATXID) (*types.ATX, error) {
+	retries := 0
+
+	for ; retries < len(a.services); func() { retries += 1; a.lastUsedSvc = (a.lastUsedSvc + 1) % len(a.services) }() {
+		if atx, err := a.services[a.lastUsedSvc].Get(ctx, id); err == nil {
+			return atx, nil
+		}
+	}
```
fasmat: keep it simple 😉 

```suggestion
	for retries := 0; retries < len(a.services); retries++ {
		if atx, err := a.services[a.lastUsedSvc].Get(ctx, id); err == nil {
			return atx, nil
		}
		a.lastUsedSvc = (a.lastUsedSvc + 1) % len(a.services)
	}
```

gateway/smesher/smesher_service_client.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (c *smesherServiceClient) GetPostConfig(ctx context.Context) (*types.PostConfig, error) {
+	resp, err := c.client.PostConfig(ctx, &emptypb.Empty{})
+	if err != nil {
+		return nil, err
+	}
+	if resp == nil {
+		return nil, errors.New("post config not found")
+	}
```
fasmat: I think `resp` is guaranteed to not be `nil` if `err` is `nil`.

gateway/smesher/smesher_service_client.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, gtw := range gateways {
+		client := newClient(gtw)
+		if cfg, err := client.GetPostConfig(ctx); err == nil {
+			return cfg
+		} else {
+			if ctx.Err() != nil {
+				return nil
+			}
+			log.With().Debug("failed to fetch post config", log.String("gateway", gtw.Target()))
+		}
```
fasmat: The `else` can be inlined.

```suggestion
		if ctx.Err() != nil {
			return nil
		}
		log.With().Debug("failed to fetch post config", log.String("gateway", gtw.Target()))
	}
```

fasmat (changes requested):

rpc/rpcserver.go
poszu: You are right. Fixed.

service/service.go
poszu: I plan to fix this problem when I rewrite Poet Service to an actor. I will hopefully get rid of all the locks and use channels only to pass data around (set new broadcaster, return and broadcast round result etc.).

service/service.go
poszu: I don't know what was the reason (I didn't write it). @moshababo do you know?

gateway/gateway_manager.go
poszu: I created a ticket and linked it in the comment. (https://github.com/spacemeshos/poet/issues/154).

Makefile
poszu: It would work too. I don't think that the performance of cloning will be an issue here. It is cloned only once after all.

For me, a git submodule is a natural tool for this kind of problem. I'm curious about what others think. @countvonzero @dshulyak  @moshababo?

service/service.go
poszu: No, it will be used for the hash (next PR).

service/service.go
poszu: `atxProvider` doesn't provide challenges, but ATXes for a submitted challenge. It's perfectly fine if a `postConfig` comes from a different Gateway than the ATXes fetched from an atxProvider. Anyway, I plan to fix all data races when I rewrite Poet Service to an actor.

gateway/activation/errors.go
poszu: Removed `Is()` implementation as not required.

Makefile
poszu: Changed to depend on the `api.h` instead.

Makefile
poszu: @fasmat Let's close on this ASAP. IMO it's not worth delaying the merge because of this. Are you OK to merge it with the git submodule and discuss it offline? Or I can change to the approach go-spacemesh took (to just put Makefile.Inc in the repo). I'm fine with either option.

types/types.go
poszu: I agree that types which are essentially part of the protocol like the ATX could have a language-agnostic schema definition in a separate repo for the sake of parallel implementations. The same for the types in `shared/types.go`.

However, the types in this file are specific for this poet implementation and are not meant to be used elsewhere.

Makefile
fasmat: lets use the copy approach; how to compile poet should be in poet not in gpu-post

service/service.go
fasmat: Lets use `github.com/spacemeshos/ed25519` everywhere in poet. The only places where `crypto/ed25519` is used is here and in `state.go` (via the `golang.org/x/crypto/ed25519` alias) and in both files it doesn't matter which ed25519 library we use.

service/validation_test.go
fasmat: @poszu what's your opinion on replacing the generic type with a specific?

Makefile
poszu: Will do. However, I think that how to build **post** should be handled by post so that the code that depends on it doesn't need to know how to build the C library that post depends on.

service/validation_test.go
poszu: Yes, it's currently used for the submitted challenge only, but it doesn't hurt IMO. I'd leave it as it is though as it is already present.

service/validation.go
```diff
@@ -0,0 +1,66 @@
...
+	// Verify against previous ATX
+	{
+		atx, err := atxs.Get(ctx, challenge.PreviousATXId)
+		if err != nil {
+			return fmt.Errorf("validation: failed to fetch previous ATX (%w)", err)
+		}
+		if !bytes.Equal(nodeID, atx.NodeID) {
+			return fmt.Errorf("validation: NodeID mismatch: challenge: %X, previous ATX: %X", nodeID, atx.NodeID)
+		}
+		if challenge.PubLayerId != atx.PubLayerID+1 {
```
poszu, line 51: It's wrong. Challenge's pub layer ID should be at least (or exactly, I'm not sure) layers-per-epoch greater than the previous ATX's pub layer ID. Similarly for challenge's pub layer ID vs positioning ATX's pub layer ID.

@moshababo could you please confirm?

service/validation.go
```diff
@@ -0,0 +1,66 @@
...
+		if meta.K1 != postConfig.K1 {
+			return fmt.Errorf("validation: K1 mismatch: challenge: %d, config: %d", meta.K1, postConfig.K1)
+		}
+		if meta.K2 != postConfig.K2 {
+			return fmt.Errorf("validation: K2 mismatch: challenge: %d, config: %d", meta.K2, postConfig.K2)
+		}
+		if meta.NumUnits < postConfig.MinNumUnits || meta.NumUnits > postConfig.MaxNumUnits {
+			return fmt.Errorf("validation: NumUnits not in the required range [%d <= %d <= %d]", postConfig.MinNumUnits, meta.NumUnits, postConfig.MaxNumUnits)
+		}
+		return proofVerifier(&initialPost.Proof, &initialPost.Metadata)
```
moshababo, line 40: The proof is being verified upon `initialPost.metadata.challenge`, but there's no check that this challenge is associated with the sender. 

Reminder: `challenge = sha256( nodeID ++ commitmentATX )`

Currently, `nodeID` is verified against prev ATX's `nodeID`, but this isn't sufficient for the initial PoST.

service/validation.go
moshababo: According to `SyntacticallyValidateAtx` (in `go-spacemesh`):
1. Challenge's `pubLayerID` must be at a higher epoch than the prev ATX's `pubLayerID`.
2. Challenge's `pubLayerID` must be at least layers-per-epoch greater than the positioning ATX's `pubLayerID`.

service/validation.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	// Verify against previous ATX
+	{
```
moshababo, line 43: This should be an `else` clause, no? 
The sender provides `initialPost` _or_ `PreviousATXId`.

poszu: It doesn't need to be an `else` because it will return on line 40 if `initialPost` is there.

poszu:
Superseded by #159

Issue #142, opened 2022-11-07
Poet potentially spins in a busy loop and generates empty proofs.


poszu:
## Affected code
The `for {}` loop in Poet `Service`.  When the condition https://github.com/spacemeshos/poet/blob/cc94dd7638f3cf546b9eebec4326266bcf49baaa/service/service.go#L202-L205 is met, it will enter a busy loop without any sleep time. 

What is more, it might generate "empty" proofs. An empty open round is not started. Poet waits for at least one submission. Once it gets one, Poet starts executing a round which might end immediately because `end` time is calculated statically for the epoch that round is created for. This `end` time might already be **in the past** the moment a round is started.

I think that it might be the reason for #127.

PR #143, opened 2022-11-07
Fix waiting for round start or ctx cancellation

poszu:
Also created https://github.com/spacemeshos/poet/issues/142 to track the busy-loop issue.

Note: the busy loop problem was already existing in the code (this PR doesn't introduce it).

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -177,38 +177,38 @@ func (s *Service) loop(ctx context.Context) {
 	var executingRounds errgroup.Group
 	defer executingRounds.Wait()
 
-	timer := time.NewTimer(0)
-	<-timer.C
-	defer timer.Stop()
 	for {
-		start := s.genesis.
-			Add(s.cfg.EpochDuration * time.Duration(s.openRound.Epoch())).
-			Add(s.cfg.PhaseShift)
-		if d := time.Until(start); d > 0 {
-			log.Info("Round %v waiting for execution to start for %v", s.openRoundID(), d)
-			timer.Reset(d)
-			select {
-			case <-timer.C:
-			case <-ctx.Done():
-				log.Info("service shutting down")
-				s.openRoundMutex.Lock()
-				s.openRound = nil
-				s.openRoundMutex.Unlock()
-				return
-			}
+		s.openRoundMutex.RLock()
+		epoch := s.openRound.Epoch()
+		s.openRoundMutex.RUnlock()
+
+		start := s.genesis.Add(s.cfg.EpochDuration * time.Duration(epoch)).Add(s.cfg.PhaseShift)
+		waitTime := time.Until(start)
+		if waitTime > 0 {
+			log.Info("Round %v waiting for execution to start for %v", s.openRoundID(), waitTime)
+		}
+		select {
+		case <-time.After(waitTime):
+		case <-ctx.Done():
+			log.Info("service shutting down")
+			s.openRoundMutex.Lock()
+			s.openRound = nil
+			s.openRoundMutex.Unlock()
+			return
 		}
 
 		s.openRoundMutex.Lock()
 		if s.openRound.isEmpty() && !s.cfg.ExecuteEmpty {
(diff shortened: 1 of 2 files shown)


service/service.go
```diff
@@ -177,38 +177,38 @@ func (s *Service) loop(ctx context.Context) {
...
+		epoch := s.openRound.Epoch()
+		s.openRoundMutex.RUnlock()
+
+		start := s.genesis.Add(s.cfg.EpochDuration * time.Duration(epoch)).Add(s.cfg.PhaseShift)
+		waitTime := time.Until(start)
+		if waitTime > 0 {
+			log.Info("Round %v waiting for execution to start for %v", s.openRoundID(), waitTime)
+		}
+		select {
+		case <-time.After(waitTime):
```
fasmat: This won't trigger at `start` but probably some time slightly earlier.

fasmat (changes requested):
Converting between a point in time and durations will continue to cause slight shifts in when certain events happen.

The problematic code is `waitTime := time.Until(start)` where `waitTime` is not exactly the time between now and `start` (conversion between monotonic and real time).

poszu:
> Converting between a point in time and durations will continue to cause slight shifts in when certain events happen.
> 
> The problematic code is `waitTime := time.Until(start)` where `waitTime` is not exactly the time between now and `start` (conversion between monotonic and real time).

These shifts don't add up. An average expected shift is statistically 0. It doesn't matter if the round starts 1ms before or after as poet time is not synchronized with anything.

service/service.go
poszu: It doesn't really matter if it triggers 1ms before. What's the right way to write this code in your opinion?

service/service.go
```diff
@@ -177,38 +177,38 @@ func (s *Service) loop(ctx context.Context) {
...
+		case <-ctx.Done():
+			log.Info("service shutting down")
+			s.openRoundMutex.Lock()
+			s.openRound = nil
+			s.openRoundMutex.Unlock()
+			return
```
fasmat: There is a leaking go routine here:

```suggestion
		timer := time.NewTimer(time.Until(s.genesis.Add(s.cfg.EpochDuration * time.Duration(epoch)).Add(s.cfg.PhaseShift))
		select {
		case <-timer.C:
		case <-ctx.Done():
			log.Info("service shutting down")
			if !timer.Stop() {
				<-timer.C
			}
			s.openRoundMutex.Lock()
			s.openRound = nil
			s.openRoundMutex.Unlock()
			return
```

service/service.go
fasmat: I'm not sure there is a right way to achieve this.

We will always have the issue that after this case triggers, `time.Now()` is not the same value as `start` but slightly different. Here in this specific case waitTime is calculated before it is used and some code executes in between.

I made a code suggestion on how to at least avoid `waitTime` to be calculated before it is used and how to avoid having dead go-routines because of `time.After`. See below 🙂

service/service.go
poszu: The proposed code:
```go
timer := time.NewTimer(time.Until(s.genesis.Add(s.cfg.EpochDuration * time.Duration(epoch)).Add(s.cfg.PhaseShift))
```

has exactly the same issue as the original. The calculations are just squeezed into one line. Avoiding a local variable `waitTime` doesn't prevent the problem. 

Where is the leaking goroutine? If you refer to garbage collection of the `Timer`  inside of `time.After()` - it is a non-issue. It is GCed after the timer hits. The only "leak" is when the application closes before the timer ticks, but it's not a problem as the app closes anyway.

fasmat: I am aware that the one liner is no different than calculating `waitTime` and then starting the timer.

The point is not to calculate `waitTime`, then doing other stuff (logging in this case) and then starting the timer. Instead sttart the timer and then log.

poszu: Done

fasmat (approved):

PR #153, opened 2022-11-14
Bump grpc-gateway to V2

poszu:
Bumped grpc-gateway to V2. 

Additionally, moved to the recommended way of tracking go tools dependencies via go.mod. See https://github.com/grpc-ecosystem/grpc-gateway#installation for details.


tools.go
```diff
@@ -0,0 +1,11 @@
+//go:build tools
+// +build tools
```
fasmat: I don't think you need both comments, `go:build` should be enough. `+build` is just for compatibility with pre-`go 1.17` versions. For details see: https://golang.org/doc/go1.18#go-build-lines

fasmat (approved):
LGTM

tools.go
poszu:
```suggestion
```

PR #158, opened 2022-11-22
Use spacemesh ed25519 instead of stdlib

fasmat:
PoET is the only repository where we are using the standard library implementation for ed25519 in some places. This PR fixes this by replacing the few uses of the library with the spacemesh implementation of ed25519.


dshulyak:
spacemesh version is much less efficient compared to stdlib, unless poet needs ed++ it doesn't make sense to use it. i am actually not sure if poet needs ed25519 library if it won't be verifying signatures

fasmat:
PoET uses both versions at the moment for:

- generating keys (stdlib & spacemesh)
- serializing them to bytes (stdlib)
- deserializing them from bytes (stdlib)

All of which make no difference in which flavour of implementation we use. Additionally ed25519 is used to:

- generate signatures for key extraction
- extract keys from signatures

both of which ONLY work with spacemeshos version of ed25519. It makes no sense to use both libraries, and since we need functionality that is not present in the stdlib we should just use the spacemesh version of the library.

pigmej:
@fasmat what are performance differences for required functions between stdlib and our implementation?

Poet really needs to be efficient as possible.

fasmat:
@pigmej we are not using ed25519 in `poet` any more. It calls a node via grpc and gets validation from there.

PR #159, opened 2022-11-22
Poet challenge verification with verifying svc

### LINKED ISSUE #146: validate submitted challenge

### LINKED ISSUE #148: implement hashing of submitted challenge

poszu:
A different approach to submitted challenge verification was proposed by @noamnelke. Poet will use a verification service exposed by its trusted gateway in the `GatewayService`.

Part of https://github.com/spacemeshos/go-spacemesh/issues/3458.
Closes #146
Closes #148 

💡 I will squash-merge to avoid mess on the develop branch.


go.mod
```diff
@@ -3,18 +3,20 @@ module github.com/spacemeshos/poet
...
+	github.com/golang/mock v1.6.0
+	github.com/google/uuid v1.1.2
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.14.0
+	github.com/hashicorp/golang-lru v0.5.4
 	github.com/jessevdk/go-flags v1.5.0
 	github.com/minio/sha256-simd v1.0.0
 	github.com/nullstyle/go-xdr v0.0.0-20180726165426-f4c839f75077
-	github.com/spacemeshos/api/release/go v1.5.1
-	github.com/spacemeshos/ed25519 v0.1.0
+	github.com/spacemeshos/api/release/go v1.5.2-0.20221124071208-887b4120e207
```
fasmat: Don't forget to first release v1.5.2 (or v1.6.0 if there are non-backwards compatible changes) ;-)

server/server.go
```diff
@@ -7,17 +7,21 @@ import (
 	"os"
 	"time"
 
+	"github.com/google/uuid"
 	proxy "github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/spacemeshos/smutil/log"
+	"go.uber.org/zap"
 	"golang.org/x/net/context"
```
fasmat:
```suggestion
	"context"
```

This isn't the only place in poet where the outdated package is used. Imo we should use `context` instead (`golang.org/x/net/context` today is just an alias to `context`)

logging/log.go
```diff
@@ -0,0 +1,20 @@
+package logging
+
+import (
+	"context"
+
+	"github.com/spacemeshos/smutil/log"
```
fasmat, line 6: Packing an instance of a logger into `context` I'm OK with, but any chance we can get rid of this dependency and use `zap.Logger` directly?

I don't really see the need to wrap a logger into a separate "util" package, especially since this package isn't concurrency safe: `log.AppLog` can be modified and read from without protection - compare with https://github.com/spacemeshos/go-spacemesh/blob/develop/log/log.go

logging/log.go
```diff
@@ -0,0 +1,20 @@
...
+
+func NewContext(ctx context.Context, logger log.Log) context.Context {
+	return context.WithValue(ctx, LoggerKey{}, logger)
+}
+
+func FromContext(ctx context.Context) log.Log {
+	if logger, ok := ctx.Value(LoggerKey{}).(log.Log); ok {
+		return logger
+	}
+	return log.AppLog
```
fasmat, line 19: imo this shouldn't fall back on a default logger but return `nil`. Otherwise we end up with a log where some logs are properly annotated and others are not; some logs are written to os.Stdout, some to os.Stderr, aso.

service/service_test.go
```diff
@@ -316,6 +330,54 @@ func TestNewService(t *testing.T) {
...
+	return (*[]byte)(&b)
+}
+
+func (Bytes) PubKey() []byte {
+	return []byte("pubkey")
+}
+
+func (Bytes) Signature() []byte {
+	return []byte("signature")
+}
```
fasmat: This can be deleted (not used anywhere)

service/service_test.go
```diff
@@ -316,6 +330,54 @@ func TestNewService(t *testing.T) {
...
+
```
poszu:
```suggestion
```

gateway/challenge_verifier/verifier.go
```diff
@@ -0,0 +1,138 @@
+package challenge_verifier
```
fasmat, line 1: The package name is a bit odd and leads to long names for functions:

`challenge_verifier.NewRoundRobinChallengeVerifier`

Suggestion: have one package per verifier and name the package according to what they do, e.g.:

`roundRobin.NewVerifier`

or if you want to keep them all in the same package:

`challengeVerifier.NewRoundRobin`

gateway/challenge_verifier/verifier.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"context"
+	"crypto/sha256"
```
fasmat: We should use the `minio` library over the standard library version:

```suggestion
	"github.com/minio/sha256-simd"
```

types/types.go
```diff
@@ -0,0 +1,16 @@
...
+//go:generate mockgen -package mocks -destination mocks/challenge_verifier.go . ChallengeVerifier
+
+type ChallengeVerificationResult struct {
+	Hash   []byte
+	NodeId []byte
+}
+
+type ChallengeVerifier interface {
+	Verify(ctx context.Context, challenge, signature []byte) (*ChallengeVerificationResult, error)
+}
```
fasmat: Imo this should be part of the `challenge_verifier` package.

types/errors.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"errors"
+)
+
+var (
+	ErrChallengeInvalid = errors.New("challenge is invalid")
+	ErrCouldNotVerify   = errors.New("could not verify the challenge")
+
+	ErrChallengeAlreadySubmitted = errors.New("challenge is already submitted")
+)
```
fasmat: Imo these should be part of the `challenge_verifier` package since that's where they are returned.

gateway/gateway_manager.go
```diff
@@ -0,0 +1,136 @@
...
+// Its Close() must be called when the connections are no longer needed.
+type Manager struct {
+	connections []*grpc.ClientConn
+}
+
+func (m *Manager) Connections() []*grpc.ClientConn {
+	if m == nil {
+		return []*grpc.ClientConn{}
+	}
+	return m.connections
```
fasmat: Is there any use case where the check for `nil` comes up? I would assume that if NewManager returns a non-`nil` `error` I cannot use the returned Manager. So we don't have to be more careful than necessary.

Also a caller of this function can modify the connections of the manager. If you want to prevent that you have to return a copy:

```suggestion
	con := make([]*grpc.ClientConn, len(m.connections))
	copy(con, m.connections)
	return con
```

gateway/gateway_manager.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (m *Manager) Close() {
+	if m == nil {
+		return
+	}
+	for _, conn := range m.connections {
+		conn.Close()
```
fasmat: `Close()` can return an `error`. Should it be propagated or handled here?

gateway/gateway_manager.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+// NewManager creates a gateway manager connected to `gateways`.
+// Close() must be called when the connections are no longer needed.
+func NewManager(ctx context.Context, gateways []string, minSuccesfullConns uint) (*Manager, error) {
+	ctx, cancel := context.WithTimeout(ctx, DefaultConnTimeout)
+	defer cancel()
```
fasmat: This overwrites a timeout set by the caller if it is longer than 10 seconds. I would find this surprising, since it is not documented that if `NewManager` fails to connect to all gateways within 10 seconds it will return with an error.

Imo we should let the caller decide about the timeout, if none is set then `NewManager` should have infinite time to connect to all gateways.

gateway/gateway_manager.go
```diff
@@ -0,0 +1,136 @@
...
+	connections, errs := connect(ctx, gateways)
+	if len(connections) < int(minSuccesfullConns) {
+		for _, conn := range connections {
+			conn.Close()
```
fasmat: A possible error isn't handled or propagated.

rpc/rpcserver.go
```diff
@@ -1,32 +1,41 @@
...
+	"github.com/spacemeshos/smutil/log"
 	"golang.org/x/net/context"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
-	"github.com/spacemeshos/poet/broadcaster"
+	"github.com/spacemeshos/poet/gateway"
+	"github.com/spacemeshos/poet/gateway/broadcaster"
+	"github.com/spacemeshos/poet/logging"
 	rpcapi "github.com/spacemeshos/poet/release/proto/go/rpc/api"
```
fasmat: the custom import name can now be dropped 🙂

rpc/rpcserver.go
```diff
@@ -47,19 +56,32 @@ func (r *rpcServer) Start(ctx context.Context, in *rpcapi.StartRequest) (*rpcapi
 		broadcastAcks = 1
 	}
 
+	gtwManager, err := gateway.NewManager(ctx, in.GatewayAddresses, connAcks)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { gtwManager.Close() }()
```
fasmat:
```suggestion
	defer gtwManager.Close()
```

rpc/rpcserver.go
```diff
@@ -83,35 +105,54 @@ func (r *rpcServer) UpdateGateway(ctx context.Context, in *rpcapi.UpdateGatewayR
...
 
+	verifier, err := service.CreateChallengeVerifier(gtwManager.Connections())
+	if err != nil {
+		return nil, fmt.Errorf("failed to create ATX provider: %w", err)
+	}
+
+	// Close the old connections and save the new manager.
+	// The temporary manager is nil-ed to avoid closing connections in defer.
+	r.gtwManager.Close()
+	r.gtwManager, gtwManager = gtwManager, nil
```
fasmat: This feels hacky. Now I also understand where the `nil` checks in the methods for `gtwManager` come from.

I feel the cleaner way would be to not use `defer` in this case but calling close directly when returning early with an error on line 120 or line 125.

rpc/rpcserver.go
```diff
@@ -83,35 +105,54 @@ func (r *rpcServer) UpdateGateway(ctx context.Context, in *rpcapi.UpdateGatewayR
...
+			return nil, status.Error(codes.FailedPrecondition, "cannot submit a challenge because poet service is not started")
+		}
+		if errors.Is(err, types.ErrChallengeInvalid) {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		} else if errors.Is(err, types.ErrCouldNotVerify) {
+			return nil, status.Error(codes.Unavailable, "failed to verify the challenge, consider retrying")
+		}
+		logging.FromContext(ctx).With().Warning("unknown error during challenge validation", log.Err(err))
+		return nil, status.Error(codes.Internal, "unknown error during challenge validation")
 	}
```
fasmat: You can use a switch / case here:

```golang
	switch {
	case errors.Is(err, service.ErrNotStarted):
		return nil, status.Error(codes.FailedPrecondition, "cannot submit a challenge because poet service is not started")
	case errors.Is(err, types.ErrChallengeInvalid):
		return nil, status.Error(codes.InvalidArgument, err.Error())
	case errors.Is(err, types.ErrCouldNotVerify):
		return nil, status.Error(codes.Unavailable, "failed to verify the challenge, consider retrying")
	case err != nil:
		logging.FromContext(ctx).With().Warning("unknown error during challenge validation", log.Err(err))
		return nil, status.Error(codes.Internal, "unknown error during challenge validation")
	}
```

service/service.go
```diff
@@ -3,6 +3,8 @@ package service
 import (
 	"bytes"
 	"context"
+	"crypto/ed25519"
```
fasmat, line 6: Since we are now switching to only using `crypto/ed25519` can you also update the import in `service/state.go`? 🙂

service/service.go
```diff
@@ -389,20 +402,52 @@ func (s *Service) executeRound(ctx context.Context, r *round) error {
...
+		err = r.submit(result.NodeId, result.Hash)
+		s.openRoundMutex.Unlock()
+		if err != nil {
+			if errors.Is(err, types.ErrChallengeAlreadySubmitted) {
+				return r, result.Hash, nil
+			}
+			return nil, nil, err
+		}
+		return r, result.Hash, nil
+	} else {
```
fasmat, line 428: you can inline the else statement

service/service.go
```diff
@@ -389,20 +402,52 @@ func (s *Service) executeRound(ctx context.Context, r *round) error {
...
+		r := s.openRound
+		err = r.submit(result.NodeId, result.Hash)
+		s.openRoundMutex.Unlock()
+		if err != nil {
+			if errors.Is(err, types.ErrChallengeAlreadySubmitted) {
+				return r, result.Hash, nil
+			}
+			return nil, nil, err
+		}
+		return r, result.Hash, nil
```
fasmat:
```suggestion
		switch {
		case errors.Is(err, types.ErrChallengeAlreadySubmitted):
			return r, result.Hash, nil
		case err != nil:
			return nil, nil, err
		}
		return r, result.Hash, nil
```

service/service.go
```diff
@@ -389,20 +402,52 @@ func (s *Service) executeRound(ctx context.Context, r *round) error {
...
+		r := s.openRound
+		err := r.submit(challenge, challenge)
+		s.openRoundMutex.Unlock()
+		if err != nil {
+			if errors.Is(err, types.ErrChallengeAlreadySubmitted) {
+				return r, challenge, nil
+			}
+			return nil, nil, err
+		}
+		return r, challenge, nil
```
fasmat:
```suggestion
		switch {
		case errors.Is(err, types.ErrChallengeAlreadySubmitted):
			return r, challenge, nil
		case err != nil:
			return nil, nil, err
		}
		return r, challenge, nil
```

gateway/broadcaster/broadcaster.go
```diff
@@ -0,0 +1,123 @@
...
+	// Count successful responses and concatenate errors of non-successful ones.
+	var numAcks int
+	var retErr error
+	for i := range b.clients {
+		res := responses[i]
+		err := errs[i]
+
+		if err != nil {
+			err = fmt.Errorf("failed to broadcast via gateway node at \"%v\" after %v: %v",
+				b.connections[i].Target(), elapsed, err)
```
fasmat:
```suggestion
			err = fmt.Errorf("failed to broadcast via gateway node at \"%v\" after %v: %w",
				b.connections[i].Target(), elapsed, err)
```

gateway/broadcaster/broadcaster.go
```diff
@@ -0,0 +1,123 @@
...
+	pbMsg := &pb.BroadcastPoetRequest{Data: msg}
+	ctx, cancel := context.WithTimeout(context.Background(), b.broadcastTimeout)
+	defer cancel()
+
+	responses := make([]*pb.BroadcastPoetResponse, len(b.clients))
+	errs := make([]error, len(b.clients))
+
+	start := time.Now()
+	var wg sync.WaitGroup
+	wg.Add(len(b.clients))
```
fasmat, line 74: This code is similar to the one used for connecting in the gateway manager. I prefer this approach with `WaitGroup` over `errGroup` and channels; it seems more straight forward to me.

Would it be much effort to adapt the other code to also use a wait group? 🙂

gateway/broadcaster/broadcaster.go
```diff
@@ -0,0 +1,123 @@
...
+			// Valid response.
+			numAcks++
+			continue
+		}
+
+		if retErr == nil {
+			retErr = err
+		} else {
+			retErr = fmt.Errorf("%v | %v", retErr, err)
+		}
```
fasmat: Maybe `retErr` should also be a type that wraps `[]error`?

server/server.go
poszu: Do you know if there is a linter to detect such problems?

logging/log.go
poszu: I'm for getting rid of the smutil wrapper. However, not this PR. I created an issue to do it later: #161.

logging/log.go
poszu: I'd rather not fallback to `nil` because all users will have to check against `nil` and fall back to something, risking a segfault. That something will probably be a default logger anyway, so I return it already here. An acceptable, but IMO less useful alternative for our use case is to return a _noop_ logger. The OpenTelemetry go lib does exactly that: https://github.com/open-telemetry/opentelemetry-go/blob/2fe8861a24e20088c065b116089862caf9e3cd8b/trace/context.go#L48-L56

gateway/gateway_manager.go
poszu: How would you handle it? IMO it could only be logged

gateway/gateway_manager.go
poszu: How would you handle it?

rpc/rpcserver.go
poszu: These two versions do different things. According to https://go.dev/doc/effective_go#defer:
> The arguments to the deferred function (which include the receiver if the function is a method) are evaluated when the defer executes, not when the call executes.

I want to evaluate the _receiver_ later because it might be overwritten:
```go
	// Close the old connections and save the new manager.
	// The temporary manager is nil-ed to avoid closing connections in defer.
	r.gtwManager.Close()
	r.gtwManager, gtwManager = gtwManager, nil
```

Now when I look at it again, the snippet could also be just:
```go
	// Swap the new and old gateway managers. 
	// The old one will be closed in defer.
	r.gtwManager, gtwManager = gtwManager, r.gtwManager
```

rpc/rpcserver.go
poszu: I realized it can be improved so that the nil checks can be removed:
```go
// Swap the new and old gateway managers. 
// The old one will be closed in defer. 
r.gtwManager, gtwManager = gtwManager, r.gtwManager
```

rpc/rpcserver.go
poszu: I didn't know you can switch on `bool`s like that in Go. Thanks!

service/service.go
poszu: It will be removed entirely in #147

types/errors.go
poszu: They are here because they are part of the interface that is also defined in the `types` package.

gateway/broadcaster/broadcaster.go
poszu: I generally prefer error groups as they are more robust (there's no need to remember to call `.Add()` and `.Done()`).

I don't agree that the broadcaster's approach is more straightforward. Instead of consumers with a clear purpose, there is the nasty loop that splits results into OKs and errors, which, IMO, is bug-prone.

types/types.go
poszu: I think this location is correct. I separated the `interface` definition from **one of** its implementations on purpose. Here is the spec of an interface that something that wants to be a challenge verifier must implement. The code in gateway/challenge_verifier is an implementation that uses a trusted gateway. The other one is the mocked implementation. In future, there might even be an in-poet implementation (not using a gateway) and the location of this interface will still be valid.

What do you think?

gateway/challenge_verifier/verifier.go
poszu: I went with the last option.

gateway/gateway_manager.go
poszu: Changed. I now return errors collected into a type wrapping `[]error`.

gateway/gateway_manager.go
poszu: Changed. I now return errors collected into a type wrapping []error.

server/server.go
fasmat: Yes, there is! `depguard` can be used for this and it is part of `golangci-lint`. See this PR https://github.com/spacemeshos/go-spacemesh/pull/3800. In it I enable it and fix some additional findings it reported.

EDIT: PR for `poet` can be found here https://github.com/spacemeshos/poet/pull/163

types/errors.go
```diff
@@ -10,3 +11,26 @@ var (
...
+	// concatenate errors
+	err := ""
+	for _, e := range e.errors {
+		if err == "" {
+			err = e.Error()
+		} else {
+			err = fmt.Sprintf("%v | %v", err, e)
+		}
+	}
+	return err
```
fasmat: You can use a `strings.Builder` for this 🙂 

```suggestion
	var b strings.Builder
	b.WriteString(e.errors[0])
	for _, e := range e.errors[1:] {
		b.WriteString(" | ")
		b.WriteString(e)
	}
	return b.String()
```

rpc/rpcserver.go
```diff
@@ -125,44 +139,42 @@ func (r *rpcServer) UpdateGateway(ctx context.Context, in *rpcapi.UpdateGatewayR
 		return nil, fmt.Errorf("failed to create ATX provider: %w", err)
 	}
 
-	// Close the old connections and save the new manager.
-	// The temporary manager is nil-ed to avoid closing connections in defer.
-	r.gtwManager.Close()
-	r.gtwManager, gtwManager = gtwManager, nil
+	// Swap the new and old gateway managers.
+	// The old one will be closed in defer.
```
fasmat, line 143: I can see this causing issues in the future if someone isn't aware that the `defer` statement is executed on the new OR the old `gtwManager` depending on when it is executed.

There are only two returns between `defer` and here. I would prefer calling `Close()` explicitly before these returns and here call close on the old `gtwManager`.

It's even fever lines of code (3 x `gtwManager.Close()`) instead of 5 lines with the `defer` statement. Let's not be more clever with the code than we need to be 🙂

rpc/rpcserver.go
```diff
@@ -76,18 +85,17 @@ func (r *rpcServer) Start(ctx context.Context, in *rpcapi.StartRequest) (*rpcapi
...
```
fasmat, line 89: See my previous comment.

gateway/broadcaster/broadcaster.go
```diff
@@ -99,25 +101,20 @@ func (b *Broadcaster) BroadcastProof(msg []byte, roundID string, members [][]byt
...
-		if retErr == nil {
-			retErr = err
-		} else {
-			retErr = fmt.Errorf("%v | %v", retErr, err)
-		}
+	retErr := types.NewMultiError(errors)
+	// If some requests failed, log it.
+	if numErrors := len(errors); numErrors > 0 {
+		log.Warning("Round %v proof broadcast failed on %d/%d gateway nodes: %v", roundID, numErrors, len(b.clients), retErr)
 	}
```
fasmat:
```suggestion
	// If some requests failed, log it.
	if len(errors) > 0 {
		log.Warning("Round %v proof broadcast failed on %d/%d gateway nodes: %v", roundID, len(errors), len(b.clients), retErr)
	}
```

go.mod
fasmat: Have you been able to `go get` v1.5.2 of the api?

types/types.go
fasmat: I'm generally against `types` packages. Imo they shouldn't exist. Either the interface should be at the users side, or at the implementers side, depending on who has the authority over it.

With that I mean: why does this interface exist?
- So it can be mocked? -> Then it should be at the users side.
- So it can be used for hiding implementation specifics behind a general interface (what imo is happening here)? Then it should be part of the implementation.

types/errors.go
fasmat: Then see my follow up comment 🙂

gateway/gateway_manager.go
fasmat: The check for `nil` was removed from this method, but not from the `Close` method. A caller of `Connections` can still modify the state of `Manager`.

rpc/rpcserver.go
fasmat: I forgot to remove this comment when I realised what is happening here. I would prefer not having a `defer` to call `gtwManager.Close()` since it makes the code unnecessarily hard to follow.

gateway/broadcaster/broadcaster.go
fasmat: I'm not sure I can follow your argument.

`errGroup.Group` has the advantage to `sync.WaitGroup` of automatically including cancellation of ongoing processes if one of the routines fails.

If you explicitly do not want to do this, the `sync.WaitGroup` seems to me a much more straight forward approach.

gateway/gateway_manager.go
poszu: Yes, `nil` check is still in `Close()`, on purpose.

Regarding returning a pointer to connections, unfortunately, Go doesn't have a notion of _pointer to const_ and I don't want to return a copy.

fasmat: The `nil`-check wouldn't be necessary if we wouldn't use `defer` to call `Close` and then possibly set the value of it to `nil` before calling it 😉

rpc/rpcserver.go
poszu: Defer prevents leaking resources, so I'd rather keep it.

gateway/gateway_manager.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	if len(errors) > 0 {
+		return types.NewMultiError(errors)
+	}
+	return nil
+}
+
+// NewManager creates a gateway manager connected to `gateways`.
+// Close() must be called when the connections are no longer needed.
+func NewManager(ctx context.Context, gateways []string, minSuccesfullConns uint) (*Manager, error) {
```
fasmat: Typo in parameter: `minSuccessfulCons`

gateway/broadcaster/broadcaster.go
poszu: My point is that `errogroup` has the advantage that I don't need to manually call `.Add()` and `.Done()` to count the running goroutines.

rpc/rpcserver.go
fasmat: That's imo a weak argument; I would even consider the use of defer dangerous here. It is not clear which gatewayManager will be closed when you `defer` and an additional `nil` check needs to be done that wouldn't be necessary if we wouldn't use `defer` to close `gatewayManager`.

gateway/broadcaster/broadcaster.go
fasmat: I do understand that, but instead it has it's own context that we don't use here and you have to return an error with every routine that is started in the `errgroup` although it will always be `nil`.

So you you are swapping `.Add()` and `.Done()` for a `return nil` and the main reason `errGroup` exists (controlled shutdown) that isn't even used.

`Package errgroup provides synchronization, error propagation, and Context cancelation for groups of goroutines working on subtasks of a common task.` vs `A WaitGroup waits for a collection of goroutines to finish.`

rpc/rpcserver.go
poszu: Manually calling `Close()` before every `return` brings a risk that in the future, somebody refactoring the code will not be aware `Close()` must be called or will forget to do it.

I don't agree it's "clever code". It's the best one can do given that Go doesn't have RAII/destructors to make sure that resources are eventually deallocated.

fasmat: There is no reason to `defer` here. Since when `defer` is called it's not even clear yet IF the method SHOULD even be called.

The `defer` sets up a call of a method on a pointer to an object that might change before the method is executed. That's the whole reason why the extra `nil` check is needed. Yes Go doesn't have destructors, so don't implement them yourself; `defer` IS NOT a substitute for a destructor, it's the Go equivalent of e.g. IDisposable from c#:

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement

types/errors.go
poszu: Except it panics when `e.errors` is empty ;)

fasmat: True, but this is never the case.

Why would you return a `MultiErr` that wraps nothing 😉

go.mod
poszu: updated

types/errors.go
poszu: Nothing prevents somebody from:
```go
MultiErr{}.Error()
```
which is a perfectly fine use of its API and would panic. It's not a matter of "why would you return a `MultiErr` that wraps noting" but providing API/implementation that's safe to use.

I will add a length check.

rpc/rpcserver.go
poszu: I don't agree that it's dangerous, nothing bad can happen. 

The additional `nil` check would need to be present anyways. Using `defer` is not the source of it. It's required because initially `r.gtwManager` might be `nil` (if there weren't enough successful gtw connections when the server was started). I can get rid of it by placing an "empty" `gateway.Manager` in this case.

gateway/broadcaster/broadcaster.go
poszu: I prefer `errgroup` because it's more robust. I don't mind that I need to return an extra `nil`, it's zero cost.

Also, `errgroup`  documentation mentions an empty group that doesn't do the cancellation so I think it is a valid use case:
> A zero Group is valid, has no limit on the number of active goroutines, and does not cancel on error.

types/types.go
poszu: Moved the interface and its errors to `gateway/challenge_verifier`.

fasmat (approved):

PR #163, opened 2022-11-26
Update linter config and fix reported issues

fasmat:
This applies the same changes as in https://github.com/spacemeshos/go-spacemesh/pull/3800 to `poet`.


.golangci.yml
```diff
@@ -56,207 +80,487 @@ output:
...
+    #   - stdmethods
+    #   - stringintconv
+    #   - structtag
+    #   - testinggoroutine
+    #   - tests
+    #   - unmarshal
+    #   - unreachable
+    #   - unsafeptr
+    #   - unusedresult
+    #   - unusedwrite
```
poszu: This list is noise only...  Available analyzers can be found as stated in the comment above:
>Run `go tool vet help` to see all analyzers.

.golangci.yml
```diff
@@ -56,207 +80,487 @@ output:
...
```
poszu: ditto

.golangci.yml
```diff
@@ -56,207 +80,487 @@ output:
...
   staticcheck:
-    # Select the Go version to target. The default is '1.13'.
-    go: "1.19"
-    # https://staticcheck.io/docs/options#checks
+    # Select the Go version to target.
+    # Default: "1.13"
+    # Deprecated: use the global `run.go` instead.
+    # go: "1.15"
+    # SAxxxx checks in https://staticcheck.io/docs/configuration/options/#checks
+    # Default: ["*"]
```
poszu: Is it really still 1.13?

poszu (commented):
IMO, the .golangci.yml became too noisy. It doesn't make sense to document every field default value, etc., as they _can change_ when the tool is updated. This information should be looked up in the docs if one is interested. Otherwise, it will get out of date pretty soon.

.golangci.yml
fasmat: This file was generated from the default config file for `golangci-lint` so I assume this is correct

.golangci.yml
fasmat: Removed default configuration options that can be found via calling a tool or from a website

.golangci.yml
fasmat: Removed default configuration options that can be found via calling a tool or from a website

poszu (approved):

PR #172, opened 2022-12-01
Refactor `Service` to actor pattern

### LINKED ISSUE #69: service.Info() accesses resources concurrently without protecting
service.go:`Info()`
reads from `openRound` and `executingRounds` while  the routine started at `Start`
writes to these variables. there's even protection (mutex) on `executingRounds` inside `Start` 
but none when reading in `Info()`.
It is important to say that `Info()` can be called at any time as it is triggered by an API call. 

cc: @moshababo

### LINKED ISSUE #140: `Service::errChan` is never read so failed rounds will block
When a round execution fails, it writes the error into `errChan`. However, nobody reads that channel anywhere. Because a write to a full channel blocks, it will lead to goroutines hanging indefinitely and leaking resources.

### LINKED ISSUE #170: Refactor `Service` to an actor
There is a lot of locking in the current implementation because the instance of the `Service` is shared and accessed from many goroutines. Readability and robustness suffer. It could be improved by refactoring to an actor pattern. Then, all state is managed in a single goroutine and any asynchronous communication with the actor happens via channels.

poszu:
With an actor pattern, all state of `Service` is mutated from a single thread. The `select` in `Service::loop` is a synchronization point.

In order to prevent accidental access to the state of `Service` from the public methods that can be called on other goroutines, a `ServiceClient` was factored out. It can talk with `Service` via a channel, sending `func(*Service)` to access/mutate `Service` safely. The sent functions are executed serially, which prevents race conditions and should be kept short and must not block.

No goroutine shares a reference to `Service` hence there is no need for locking at all. 

Closes #69 
Closes #140 
Closes #170


service/service.go
```diff
@@ -234,27 +272,36 @@ func (s *Service) Start(b Broadcaster, atxProvider challenge_verifier.Verifier)
...
-		s.newRound(ctx, uint32(epoch))
+		s.openRound = s.newRound(ctx, uint32(epoch))
 	}
-	s.openRoundMutex.Unlock()
+
+	cmds := make(chan Command, 1)
+	s.ServiceClient = &ServiceClient{
+		command: cmds,
+	}
+	s.ServiceClient.SetChallengeVerifier(atxProvider)
```
poszu: Rename `atxProvider` 🤦

service/service.go
```diff
@@ -71,24 +79,19 @@ type Service struct {
...
 
 	PubKey  ed25519.PublicKey
 	privKey ed25519.PrivateKey
 
-	broadcaster       atomic.Value // holds Broadcaster interface
-	challengeVerifier atomic.Value // holds challenge_verifier.Verifier
+	broadcaster Broadcaster
 
-	errChan chan error
 	sync.Mutex
```
countvonzero, line 93: is this still used?

service/service.go
```diff
@@ -170,59 +173,96 @@ func NewService(cfg *Config, datadir string) (*Service, error) {
...
+	if waitTime > 0 {
+		logging.FromContext(ctx).With().Info("waiting for execution to start", log.Duration("wait time", waitTime), log.String("round", round.ID))
 	}
+	return timer
 }
 
-func (s *Service) Start(b Broadcaster, atxProvider challenge_verifier.Verifier) error {
+func (s *Service) Start(b Broadcaster, verifier challenge_verifier.Verifier) error {
 	s.Lock()
 	defer s.Unlock()
```
countvonzero, line 283: can get rid of the mutex using sync.Once?

service/service.go
```diff
@@ -234,27 +274,36 @@ func (s *Service) Start(b Broadcaster, atxProvider challenge_verifier.Verifier)
 	ctx, stop := context.WithCancel(context.Background())
 	s.stop = stop
 
-	s.SetBroadcaster(b)
-	s.SetChallengeVerifier(atxProvider)
+	s.broadcaster = b
 
+	var toResume []*round
```
countvonzero, line 295: can you have multiple rounds ongoing?

service/service.go
```diff
@@ -279,168 +328,145 @@ func (s *Service) Started() bool {
...
 
-func (s *Service) SetBroadcaster(b Broadcaster) {
-	if s.broadcaster.Swap(b) != nil {
-		log.Info("Service broadcaster updated")
+func (s *ServiceClient) SetBroadcaster(b Broadcaster) {
+	s.command <- func(s *Service) {
+		old := s.broadcaster
+		s.broadcaster = b
+		if old != nil {
+			log.Info("Service broadcaster updated")
```
countvonzero, line 398: maybe more informative to log both the old and the new broadcaster?

service/service.go
```diff
@@ -279,168 +328,145 @@ func (s *Service) Started() bool {
...
+		}
 	}
 }
 
-func (s *Service) SetChallengeVerifier(provider challenge_verifier.Verifier) {
+func (s *ServiceClient) SetChallengeVerifier(provider challenge_verifier.Verifier) {
```
countvonzero, line 403: i don't understand why broadcaster is updated via the actor but not verifier

service/service.go
```diff
@@ -279,168 +328,145 @@ func (s *Service) Started() bool {
...
-	return nil
+	return roundResult{round: r, err: err}
 }
 
-func (s *Service) Submit(ctx context.Context, challenge, signature []byte) (*round, []byte, error) {
-	logger := logging.FromContext(ctx)
-	if !s.Started() {
-		return nil, nil, ErrNotStarted
+func (s *ServiceClient) Submit(ctx context.Context, challenge, signature []byte) (string, []byte, error) {
+	if s == nil {
```
countvonzero: how can we get here if s is nil? do you mean to check s.Started()?

service/service.go
```diff
@@ -279,168 +328,145 @@ func (s *Service) Started() bool {
...
-	r := s.openRound
-	err = r.submit(result.NodeId, result.Hash)
-	s.openRoundMutex.Unlock()
+
+	type response struct {
+		round string
+		err   error
+	}
+	done := make(chan response, 1)
+	s.command <- func(s *Service) {
```
countvonzero, line 428: not familiar with the app structure in poet. is there a timeout mechanism here? it could block forever if the last command processing get stuck

service/service.go
```diff
@@ -279,168 +328,145 @@ func (s *Service) Started() bool {
...
+		done <- response{
+			round: s.openRound.ID,
+			err:   s.openRound.submit(result.NodeId, result.Hash),
+		}
+		close(done)
+	}
+	resp := <-done
```
countvonzero: same here. in case the processing of the command hangs

service/service.go
```diff
@@ -170,59 +173,96 @@ func NewService(cfg *Config, datadir string) (*Service, error) {
...
-			log.Info("Round %v waiting for execution to start for %v", s.openRoundID(), waitTime)
-		}
 		select {
+		case cmd := <-cmds:
+			cmd(s)
+
+		case result := <-roundResults:
+			if result.err == nil {
+				broadcaster := s.broadcaster
+				go broadcastProof(s, result.round, result.round.execution, broadcaster)
```
countvonzero, line 235: use eg?

service/service.go
```diff
@@ -170,59 +173,96 @@ func NewService(cfg *Config, datadir string) (*Service, error) {
...
-
-		start := s.genesis.Add(s.cfg.EpochDuration * time.Duration(epoch)).Add(s.cfg.PhaseShift)
-		waitTime := time.Until(start)
-		timer := time.After(waitTime)
-		if waitTime > 0 {
-			log.Info("Round %v waiting for execution to start for %v", s.openRoundID(), waitTime)
-		}
 		select {
+		case cmd := <-cmds:
+			cmd(s)
```
countvonzero, line 230: i would do eg.Go() to execute commands. we don't know if the passed in func can block for a long time or not.

service/service.go
poszu: It is used to synchronize `Service::Start()`. It's required because `Start()` is called on RPC and hence can be many times called in parallel.

service/service.go
poszu: Probably yes. It would be explicit that `Service` cannot be restarted after `Shutdown()`. I will explore this idea in a separate PR as it is not really related to the scope of this one, OK?

service/service.go
poszu: Theoretically no - the poet should execute only 1 round at a time. However, it could be that we recover multiple ongoing rounds (the state is read from the disk and might be wrong). It's not yet clear to me how to handle such case so I preserved the current behaviour - to pick up and execute all.

service/service.go
poszu: Broadcaster is going to be removed very soon.

service/service.go
poszu: Because `verifier` is not part of `Service`. `ServiceClient` uses it to verify the challenge **before** passing it on to the `Service` for registration.

service/service.go
poszu: So the `ServiceClient` is created when the `Service` is started (it was `nil` before). However, I just realized that's a race condition. I will change it. Thanks!

service/service.go
poszu: It's a core requirement that the "commands" are **short** and **don't block**. It's not really specific to the actor pattern - it was exactly the same with locks, it was just not visible in an obvious way. With locks, a method could block on a `mutex.Lock()` for an indefinite time (until it is unlocked in a different place, if at all).

If processing a previous command gets stuck, then it is a bug that needs fixing. I can, however, do a `select` on `done` and `ctx.Done()` to let the caller a way to "give up".

service/service.go
poszu: Broadcasting is meant to be "fire and forget" - nobody will `.Wait()` for it to finish so there's no use case for a wait/error group here.

Anyway, I'm removing broadcasting altogether in the next PRs soon.

service/service.go
poszu: The command cannot be run on a goroutine as it would be racy. It's the core idea to run them _sequentially_. 

The passed functions cannot block, it's a requirement. For example, they must not do any IO.

fasmat: Could you add this documentation to the definition of `Command`?

service/service.go
```diff
@@ -71,24 +79,19 @@ type Service struct {
...
 
-	broadcaster       atomic.Value // holds Broadcaster interface
-	challengeVerifier atomic.Value // holds challenge_verifier.Verifier
+	broadcaster Broadcaster
 
-	errChan chan error
 	sync.Mutex
 }
 
+type Command func(*Service)
```
fasmat, line 99: Please add some documentation to `Command`, e.g. that the command must terminate in a timely manner and should not do file I/O.

Imo a command should also receive the context of the function that actually executes it, so it can return early if the work it does isn't required anymore.

service/service.go
fasmat: I feel like `broadcasProof` should be a method of `round` shouldn't it? I'm also with @countvonzero here.

"Fire and forget" doesn't go well with a graceful shutdown. `boradcastProof` isn't guaranteed to return in a timely manner. It retries broadcasting possibly for many times (config parameter) and has no way of returning early, e.g. when the timer of the round expires.

service/service.go
fasmat: > If processing a previous command gets stuck, then it is a bug that needs fixing. I can, however, do a select on done and ctx.Done() to let the caller a way to "give up".

I would welcome that

service/service.go
```diff
@@ -279,168 +328,145 @@ func (s *Service) Started() bool {
...
 	}
-	return r, result.Hash, nil
+	logger.With().Debug("submitted challenge for round", log.String("round", resp.round))
+	return resp.round, result.Hash, nil
 }
 
-func (s *Service) Info() (*InfoResponse, error) {
-	if !s.Started() {
+func (s *ServiceClient) Info() (*InfoResponse, error) {
+	if s == nil {
```
fasmat: Why is this check for `nil` necessary in multiple methods?

service/service_test.go
```diff
@@ -11,13 +11,12 @@ import (
...
 	"github.com/stretchr/testify/require"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/spacemeshos/poet/gateway/challenge_verifier"
 	"github.com/spacemeshos/poet/gateway/challenge_verifier/mocks"
-	"github.com/spacemeshos/poet/prover"
+	"github.com/spacemeshos/poet/service"
 )
 
 type MockBroadcaster struct {
```
fasmat, line 23: Can we use `gomock` for this?

fasmat (commented):
I'm still going through the code and trying to understand it so more comments might follow 🙂

service/service_test.go
```diff
@@ -74,130 +69,68 @@ func TestService_Recovery(t *testing.T) {
...
 	}
 
 	// Submit challenges to open round (0).
-	submitChallenges(0, 0)
-
-	// Verify that round is still open.
-	req.Equal(rounds[0].ID, s.openRoundID())
+	submitChallenges("0", challengeGroups[0])
 
 	// Wait for round 0 to start executing.
```
fasmat, line 86: The comment says waiting for round 0 to execute, but the assert checks if round 1 is open?

service/service_test.go
```diff
@@ -207,7 +140,7 @@ func TestConcurrentServiceStartAndShutdown(t *testing.T) {
 	t.Parallel()
```
fasmat, line 141: I wonder if `TestConcurrentServiceStartAndShutdown` is even a requirement for Poet?

Should a single service be used multiple times in the same running poet? Why? If an operator wants to run multiple poets on the same machine they can just run multiple instances of the program / docker container with different settings?

service/service.go
```diff
@@ -279,168 +328,145 @@ func (s *Service) Started() bool {
...
-	s.executingRounds[r.ID] = r
-	s.executingRoundsMutex.Unlock()
-
-	defer func() {
-		s.executingRoundsMutex.Lock()
-		delete(s.executingRounds, r.ID)
-		s.executingRoundsMutex.Unlock()
-	}()
-
+func executeRound(ctx context.Context, r *round, end time.Time, minMemoryLayer uint) roundResult {
```
fasmat: This should be a method of `round` imo.

There is a `round::execute()` method that is only called form here. Maybe replace that method with this function?

service/service.go
```diff
@@ -170,59 +173,96 @@ func NewService(cfg *Config, datadir string) (*Service, error) {
...
-			if err := s.executeRound(ctx, round); err != nil {
-				s.asyncError(fmt.Errorf("round %v execution error: %v", round.ID, err))
-				return nil
-			}
-			broadcastProof(s, round, round.execution, s.getBroadcaster())
-			return nil
-		})
+func (s *Service) roundEndTime(round *round) time.Time {
+	return s.roundStartTime(round).Add(s.cfg.EpochDuration).Add(-s.cfg.CycleGap)
+}
```
fasmat, line 270: I feel like these two should be methods of `round`: `round::Start()` and `round::End()`.

service/service.go
```diff
@@ -279,168 +328,145 @@ func (s *Service) Started() bool {
 	return s.started.Load()
 }
 
-func (s *Service) recover(ctx context.Context) error {
-	log.With().Info("Recovering service state", log.String("datadir", s.datadir))
+func (s *Service) recover(ctx context.Context) (open *round, executing []*round, err error) {
+	logger := log.AppLog.WithName("recovery")
+	logger.With().Info("Recovering service state", log.String("datadir", s.datadir))
 	entries, err := os.ReadDir(s.datadir)
```
fasmat, line 346: This seems like a good place to use https://pkg.go.dev/path/filepath#WalkDir. What do you think?

service/service.go
```diff
@@ -279,168 +328,145 @@ func (s *Service) Started() bool {
...
-	s.executingRoundsMutex.RUnlock()
-
-	return &InfoResponse{
-		OpenRoundID:        s.openRoundID(),
-		ExecutingRoundsIds: ids,
-	}, nil
+	return <-resp, nil
 }
 
 // newRound creates a new round with the given epoch. This method MUST be guarded by a write lock on openRoundMutex.
```
fasmat: I feel like this comment isn't accurate any more.

Also, why does this function `panic`?

service/service.go
poszu: I'm going to remove broadcaster soon in PR for #168

service/service_test.go
poszu: Broadcaster is going to be removed in #168

service/service_test.go
poszu: It's required because `Start()` and `Shutdown()` might be called on different goroutines (Start - on a GRPC call, Shutdown - when the app closes).

service/service.go
poszu: A `round` isn't aware of genesis, epoch duration, phase shift etc. so it couldn't calculate when to start and when to end.

service/service.go
poszu: Great idea. I will explore it in a separate PR refactoring `Service::recover()`. I created an issue to track it https://github.com/spacemeshos/poet/issues/179

service/service.go
fasmat: I feel like there is too much coupling between `Service` and `round`.

Suggestion: instead of specifying the `epoch` when creating a `round` with `newRound`, we could pass in the start and end time of the round.

The epoch is only needed for determining these two variables and serves as an ID for the round. If I'm not mistaken the ID of a round only needs to be unique, so `startTime` could also serve as ID for a round. What's your opinion?

service/service.go
poszu: Will update the doc, thanks. It panics if it couldn't save the state (this part is unchanged in the PR).

service/service.go
poszu: Do you mean the context of the function that creates and schedules the `Command` or the context of `Service::loop`? The former is already available as a variable captured by a closure.  If I pass the context of `Service::loop` there will be two contexts available which will be confusing.

service/service.go
poszu: See https://github.com/spacemeshos/poet/pull/172#discussion_r1043039779. Anyway - I'm changing it.

service/service_test.go
poszu: It relies on the fact that rounds IDs are sequential. Round 0 must have started execution if round 1 is open.

service/service.go
poszu: Yes, there is already `round::execute()` that has all the "execution logic". The purpose of this function is only to hide the "boilerplate code" (mostly logging) and keep the Service loop concise.

service/service.go
poszu: That's an interesting idea. @noam also pointed out recently that rounds/poet shouldn't be aware of epochs. Anyway - `round` refactoring is out of the scope of this PR. I will be refactoring `round` separately (I want to de-async its teardown logic and reduce the number of channels there).

service/service.go
fasmat: The context of the caller of Command, not the creator of it.

```golang
type Command func(ctx context.Context, *Service)

// creator
cmdChan <- func(ctx context.Context, s *Service) {
	select {
	case <- ctx.Done():
		return
	default:
	}

	s.variable = newValue
}

// caller (b.loop)
cmd := <-cmdChan
cmd(ctx, s) // ctx here is the context that cancels loop
```

service/service_test.go
fasmat: It waits for round 1 to open, not round 0 to start executing. It's confusing if a comment says something different from the code that follows.

service/service.go
fasmat: OK, then can this code be moved into the `round::execute()` method to remove this code and keep the code concise? 🙂

service/service.go
fasmat: Could you create an issue for that or add it as requirement to an existing issue?

service/service.go
fasmat: So if the state fails to persist poet just crashes silently?

service/service.go
poszu: Done: https://github.com/spacemeshos/poet/issues/181

service/service.go
```diff
@@ -433,25 +443,30 @@ func (s *ServiceClient) Submit(ctx context.Context, challenge, signature []byte)
...
+			return resp.round, result.Hash, nil
+		case err != nil:
+			return "", nil, resp.err
+		}
+		logger.With().Debug("submitted challenge for round", log.String("round", resp.round))
 		return resp.round, result.Hash, nil
-	case err != nil:
-		return "", nil, resp.err
+	case <-ctx.Done():
+		return "nil", nil, ctx.Err()
```
fasmat: Why `"nil"` and not empty string?

service/service.go
poszu: Why silently? It should emit a backtrace.

fasmat: Silently in a sense as it doesn't log why it decided to panic.

I know there will be a backtrace but that is not a structured log message that can easily be searched for e.g. in 10_000 of logs that we emit during a systest.

service/service.go
poszu: It will be confusing because it will look like this:
```go
func (s *ServiceClient) Info(ctx context.Context) (*InfoResponse, error) {
	s.command <- func(serviceCtx context.Context s *Service) {
		// THERE ARE 2 Contexts here, it's confusing
		select {
		case <-ctx.Done():
			return // Info's context is cancelled - GRPC is cancelled
		case <- serviceCtx.Done():
			return // Service's context is cancelled - shutting down
		default:
		}
		s.variable = newValue
	}
}
```

Relying on the requirement that `Command` is supposed to be **short**, I don't see a need to select on any of these. It should just assign `newValue`  / run and return.

service/service.go
poszu: ups 👍

service/service.go
fasmat: The code you posted is wrong. A `Command` should never check the context of the function that created it. That doesn't make any logical sense.

With the same argument I could say that a command shouldn't have a `Service` parameter, because it already has access to the service from its parents scope.

That being sad: I think `ServiceClient` isn't required as a type. imo inlining it into `Service` makes the code much more straight forward than having an embedded service client.

fasmat: Like this: https://github.com/spacemeshos/poet/compare/170-refactor-service-to-actor...inline-service-client

Code is still free from data races but all the nesting is removed.

service/service.go
poszu: A panic prints the message of the error that is passed in. The error should already contain the _why_ and _what_.

fasmat: The error that the panic prints will be something along the lines of `"permission denied"` or `"file already closed"` without any further context

service/service.go
poszu: The point of extracting a `ServiceClient` is to prevent accidental access to the state of `Service` without synchronization. It is just not possible and guaranteed at compile time. It is to make the code more robust and to prevent mistakes leading to data races. This guard is lost when the `Info` and `Submit` methods are integrated back into the `Service` as they can now access the state of `Service` without any synchronization.

fasmat: But what ensures that any method that changes the `Service` isn't part of `Service` but part of `ServiceClient`?

poszu: The key point of the actor pattern is that all access to the state of the actor (`Service`) is done in a single goroutine.

After this refactoring, there isn't any goroutine, other than the one running `Serivce::loop`, that has a reference to the `Service`. Thanks to this all mutations and reads of the state are synchronized. 

It's perfectly fine if there is another private method of `Service` that accesses its state as there is only one goroutine that can call them (the one running the `Service::loop()`. The `Service` must expose all its public API thru the `ServiceClient` to make the assumptions hold.

service/service.go
poszu: Resolving as the `newRound` documentation is fixed. 

@mafa The panic topic can be continued. Please create an issue if this needs to be addressed in your opinion.

service/service.go
countvonzero: reading service.go after your change, the differentiation between `Service` and `ServiceClient` is still unclear to me.
i don't understand why ServiceClient cannot be inlined.
it seems the public APIs Info/Submit/SetBroadcaster/SetChallengeVerifier all can submit through the command channel.

service/service.go
countvonzero: the issue's description is empty. why removing broadcaster? 
i think it's better to do the right thing now instead of relying on future implementation.
the point here is to manage shutdown withe errgroup and there is really no good reason not to do it here.

service/service.go
countvonzero: please file an issue for this to at least explore the correctness. 
recovering multiple rounds doesn't seem to fit within the poet design. is the code even capable of running multiple rounds at the same time?

service/service.go
countvonzero: sure. up to you. it's used often in go-spacemesh. 

> It would be explicit that Service cannot be restarted after Shutdown()

i don't see there is any way to restart it after the Shutdown() RPC call. is there?

poszu: No, there isn't even a shutdown RPC. `Shutdown()` is called on program exit.

service/service.go
poszu: The code is capable of running multiple rounds at once. In this PR I want to preserve the existing behaviour. 

Created a ticket to track it: https://github.com/spacemeshos/poet/issues/182

service/service.go
poszu: It's just a subtask auto-created from https://github.com/spacemeshos/pm/issues/173. I'm implementing it right now. There is no point in fixing it as I have already removed the broadcaster in my local changes. I will open a PR with it soon.

service/service.go
poszu: `ServiceClient` _could_ be inlined into `Service`. Not inlining it prevents a whole class of bugs as there is literally no way to access the `Service` state from `ServiceClient` in any other way than safely via the `commands` channel. Thanks to this, it is impossible to accidentally create a data race from the `Submit` and `Info` methods.

service/service.go
countvonzero: if you insist, i am ok with it.

> There is no point in fixing it

but it's literally doing `eg.Go()` instead of `go`....

service/service.go
countvonzero: i agree with the general approach to move it to the actor pattern. 

tho as is, `Service` still requires a lock that protects it from doing multiple `Start()` calls, where members of `Service` is mutated. more over, there is a `ServiceClient` that separates out data that need changed during the lifetime of this server.

`Service` and `ServiceClient` is not distinguishable from the user of this service. it serves the same purpose and have the same life-span.

i think the cleanest way to do this
- just one struct Service (get rid of ServiceClient)
- make `Start()` and `SetChallengeVerifier()` also submit the actual work to commands channel.
- remove all mutex (as the only thing that will be mutated is the commands channel)

countvonzero (approved):
approving as there are more upcoming work for the poet spam attack.

giving the author the choice of addressing feedback in this PR, or leave TODO/FIXME now and address them in the upcoming PRs.

service/service.go
countvonzero: need to start the loop when service is created too

service/service.go
poszu: It's not _that_ simple because it could block shutdown for a very long time (the default is 500 minutes) if broadcasting fails as there is no cancellation mechanism implemented in `broadcastProof()`.

service/service.go
poszu: Thanks, @countvonzero. I'm going to move the start of the event loop to where the `Service` is created, which will allow me to remove the last mutex. I'm going to do this in a PR removing broadcasting as this is the only reason for delaying the Service start (it cannot operate w/o a broadcaster).

PR #186, opened 2022-12-12
Avoid saving Service state repeatedly

poszu:
In the past, the state contained information (the next round ID) that had been changing during `Service` lifetime. It doesn't anymore so there is no point in saving it every time a new round is opened.

After this patch, the state of `Service` is only saved after it's been created for the first time.


service/service.go
```diff
@@ -165,12 +159,15 @@ func NewService(cfg *Config, datadir string) (*Service, error) {
...
-	state, err := state(datadir)
+	state, err := loadServiceState(datadir)
 	if err != nil {
 		if !errors.Is(err, ErrFileIsMissing) {
 			return nil, err
 		}
-		state = initialState()
+		state = newServiceState()
+		if err := state.save(datadir); err != nil {
+			return nil, fmt.Errorf("failed to save state: %w", err)
```
fasmat, line 169: No more panic 👍

fasmat (approved):

Issue #181, opened 2022-12-09
Refactor `round`


poszu:
Goal:
- simplify the structure and logic,
- de-async teardown procedure - it should be called by the owner of the `round` when it is no longer needed to avoid data races.

PR #187, opened 2022-12-12
Allow for querying proofs and remove broadcasting

### LINKED ISSUE #167: Persist the generated proofs

### LINKED ISSUE #168: Remove integration with gateway nodes for broadcasting proofs

### LINKED ISSUE #169: Add RPC API for querying proof results of executed rounds.
follows: #167

poszu:
Implements necessary poet changes for https://github.com/spacemeshos/pm/issues/173

Closes #167
Closes #168
Closes #169 

### Notable changes:
- `Service` reports produced proofs via a channel instead of broadcasting them. The proofs are then picked up by consumers:
  - in a poet server case, the consumer is a database actor that serializes and persists them
  - in tests, the test is a consumer directly
- removed the last `Mutex` from the `Service` struct. The actor event loop is now _synchonous_ (started via `Service::Run()`). It doesn't spawn a goroutine and finishes when the context is cancelled. It made the shutdown logic **redundant**.
- split server startup into:
  - creation: `server::New()`, where it configures itself and binds ports to listen on
  - start: `Server::Start`, where it actually starts to process incoming requests
  
  It allows writing tests that run on random ports **and obtain ports the server listens on**, which lets run parallel tests.
- `Submit()` GRPC returns duration till the end of the round that the challenge was submitted to. It gives two benefits for the node:
  - it can easily verify if the resulting proof will be valuable for it (if it ends _on time_). It might not be if it missed the registration window,
  - it knows when to query for the proof


rpc/api/api.proto
```diff
@@ -46,6 +47,15 @@ service Poet {
             get: "/v1/info"
         };
     }
+
+    /**
+    GetProof returns the generated proof for given round id.
+    */
+    rpc GetProof(GetProofRequest) returns (GetProofResponse) {
+        option (google.api.http) = {
+            get: "/v1/proof/{round_id}"
```
countvonzero: restful naming convention tends to use plural for resources. proofs

shared/shared.go
```diff
@@ -68,6 +68,25 @@ func MakeLabelFunc() func(hash func(data []byte) []byte, labelID uint64, leftSib
...
+
+	// NumLeaves is the width of the proof-generation tree.
+	NumLeaves uint64
+}
+
+type ProofMessage struct {
+	Proof
+	ServicePubKey []byte
+	RoundID       string
+	Signature     []byte
```
countvonzero: what is the intended use for Signature? (i cannot find it from the pm issue)
it seems set to nil always currently.

service/db.go
```diff
@@ -46,3 +54,62 @@ func (db *LevelDB) Delete(key []byte) error {
 func (db *LevelDB) Iterator() iterator.Iterator {
 	return db.DB.NewIterator(nil, db.ro)
 }
+
+type ProofsDatabase struct {
+	db     *leveldb.DB
```
countvonzero, line 19: i think it's more consistent to reuse LevelDB and its interface here.

service/db.go
```diff
@@ -46,3 +54,62 @@ func (db *LevelDB) Delete(key []byte) error {
...
+}
+
+func (db *ProofsDatabase) Run(ctx context.Context) error {
+	logger := log.AppLog.WithName("proofs-db")
+	for {
+		select {
+		case proof := <-db.proofs:
+			serialized, err := serializeProofMsg(proof)
+			if err != nil {
+				logger.With().Error("failed serializing proof", log.Err(err))
```
countvonzero: stop work here, or log fatal, i prefer the latter as this should not happen.

service/db.go
```diff
@@ -46,3 +54,62 @@ func (db *LevelDB) Delete(key []byte) error {
...
+			}
+			if err := db.db.Put([]byte(proof.RoundID), serialized, &opt.WriteOptions{Sync: true}); err != nil {
+				logger.With().Error("failed storing proof in DB", log.Err(err))
```
countvonzero, line 55: break select

server/server.go
```diff
@@ -20,15 +22,60 @@ import (
...
+func (s *Server) Close() {
+	s.rpcListener.Close()
+	s.restListener.Close()
+}
+
+func (s *Server) RpcAddr() net.Addr {
+	return s.rpcListener.Addr()
+}
+
 // startServer starts the RPC server.
```
countvonzero: update comment

server/server.go
```diff
@@ -20,15 +22,60 @@ import (
...
+	restListener net.Listener
+}
+
+func New(cfg config.Config) (*Server, error) {
+	rpcListener, err := net.Listen(cfg.RPCListener.Network(), cfg.RPCListener.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to listen: %v", err)
+	}
+
+	restListener, err := net.Listen(cfg.RESTListener.Network(), cfg.RESTListener.String())
```
countvonzero, line 44: sorry for lack of context. what was the original reason for not using grpc for getting proof, but using restful?

service/service.go
```diff
@@ -57,28 +48,29 @@ type serviceState struct {
 // ServiceClient is an interface for interacting with the Service actor.
 // It is created when the Service is started.
 type ServiceClient struct {
-	serviceStarted    *atomic.Bool
+	start             chan struct{}
```
countvonzero: not clear to me the diff btwn this `start` and the one in `Service`

countvonzero (commented):
some high level questions first to get the context. thanks for your patience.

rpc/api/api.proto
poszu: Does it? For example GH has:
- `/pulls` and `/issues` to list many,
- `/pull/<id>` and `/issue/<id>`  to get exactly one.

shared/shared.go
poszu: Yes, it has been like that since ever (the same structure was used for broadcasting proofs before). I think that eventually proofs are supposed to be signed. There's been a discussion around this topic recently but w/o a clear decision (sorry for a private link): https://spacemesh.slack.com/archives/C89QJJY3G/p1668892400809239.

service/db.go
poszu: I'd rather remove the LevelDB wrapper altogether as it only wraps `leveldb.DB` in order to not have to provide ro/wo...

service/db.go
poszu: I will return an error. The owner of DB will have a choice of what to do next.

server/server.go
poszu: Poet provides both "raw" GRPC and REST API via a [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) reverse proxy. The clients are free to choose which one they use.

I don't know why the client in go-spacemesh uses the HTTP REST API 🤷. I created an issue some time ago to change the client to use grpc natively: https://github.com/spacemeshos/go-spacemesh/issues/3702.

service/service.go
poszu: Refactored and removed `ServiceClient`.

shared/shared.go
poszu: Note: I just moved this structure from `service.go`. It was named `GossipPoetProof` before.

rpc/api/api.proto
countvonzero: yes. i believe that's the industry best practice. try google "restful naming conventions"
resources are queried by their id. in GH's case, it's actually confusing to the API users that both `issues` and `issue` are available and open to typo mistakes.

shared/shared.go
countvonzero: right. can you remove it for now? 
it's straightforward to add it back. creating an illusion that there is a signature is bad.

service/db.go
countvonzero: what's the problem with hiding ro/wo? 

anyway, i am fine with you removing LevelDB in this PR.
my main concern is that you are creating a new way to use level db in this repo, which make the code confusing.

poszu: >what's the problem with hiding ro/wo?

Hiding ro/wo in the wrapper struct limits all read/write operations to use the same ro/wo. It might not always be optimal and necessary.

Additionally, it requires providing a wrapper method for every method of `leveldb.DB` that needs to be used.

IMO, the benefit of NOT having to explicitly set ro/wo on each operation is illusory and can actually be harmful (hurting performance in the best case or introducing bugs in the worst case - when an improper ro/wo is used).

I planned to remove the wrapper in #181 anyway but I can do it immediately.

rpc/rpcserver.go
```diff
@@ -184,3 +160,35 @@ func (r *rpcServer) GetInfo(ctx context.Context, in *api.GetInfoRequest) (*api.G
 
 	return out, nil
 }
+
+// GetProof implements api.PoetServer.
+func (r *rpcServer) GetProof(ctx context.Context, in *api.GetProofRequest) (*api.GetProofResponse, error) {
+	if info, err := r.s.Info(ctx); err == nil {
+		if info.OpenRoundID == in.RoundId || slices.Contains(info.ExecutingRoundsIds, in.RoundId) {
```
countvonzero, line 167: i still finds this multiple rounds logic confusing. would appreciate (for future reviews) that we get rid of it if we know poet is not doing multiple rounds at the same time.

rpc/rpcserver.go
```diff
@@ -184,3 +160,35 @@ func (r *rpcServer) GetInfo(ctx context.Context, in *api.GetInfoRequest) (*api.G
...
+			return nil, status.Error(codes.Unavailable, "round is not finished yet")
+		}
+	}
+
+	proof, err := r.proofsDb.Get(ctx, in.RoundId)
+	switch {
+	case errors.Is(err, service.ErrNotFound):
+		return nil, status.Error(codes.NotFound, "proof not found")
```
countvonzero, line 175: would printing err.Error() here be more informative for debugging?

service/round.go
```diff
@@ -139,16 +143,16 @@ func (r *round) submit(key, challenge []byte) error {
...
 	}
 
-	if has, err := r.challengesDb.Has(key); err != nil {
+	if has, err := r.challengesDb.Has(key, nil); err != nil {
 		return err
 	} else if has {
 		return fmt.Errorf("%w: key: %X", ErrChallengeAlreadySubmitted, key)
 	}
-	return r.challengesDb.Put(key, challenge)
+	return r.challengesDb.Put(key, challenge, &opt.WriteOptions{Sync: true})
```
countvonzero, line 151: i have to say i do like this better (not hiding wo/ro) 👍🏽

countvonzero (approved):

rpc/rpcserver.go
poszu: The GRPC interceptor already prints requests and responses here :wink:: https://github.com/spacemeshos/poet/blob/c2c2e82236aeef1fc8b18d69bd0a6fce5bb14ff2/server/server.go#L174-L191

The returned error will be visible in logs printed there.

PR #188, opened 2022-12-12
Extend nonce space in fiat shamir

### LINKED ISSUE #173: `FiatShamir` function can enter an infinite loop
### The affected code:
https://github.com/spacemeshos/poet/blob/74fd3c58308acc7f3cce627ce8b4ce810d556c85/shared/shared.go#L25-L40

### What happens
The `FiatShamir` function might not be able to finish the following loop: https://github.com/spacemeshos/poet/blob/74fd3c58308acc7f3cce627ce8b4ce810d556c85/shared/shared.go#L34-L38

when `spaceSize` is just little bigger than `securityParam`. It might not be able to generate enough unique `id`s.

### How to reproduce
Run test:
```go
func TestFiatShamirHang(t *testing.T) {
	FiatShamir([]byte("challenge this"), uint64(T)+1, T)
}
```

poszu:
Closes #173 

Extended the nonce space in `FiatShamir` from 1 byte to 4 bytes.

### DIFF
--- a/shared/shared_test.go
+++ b/shared/shared_test.go
@@ -53,6 +53,12 @@ func TestFiatShamirReturnsCorrectNumOfIndices(t *testing.T) {
 	require.Len(t, FiatShamir(challenge, 20, 15), 15)
 }
 
+// This test used to hang.
+// See https://github.com/spacemeshos/poet/issues/173
+func TestFiatShamirLowSpaceSize(t *testing.T) {
+	FiatShamir([]byte("challenge this"), uint64(T)+1, T)
+}
+
 func TestMakeLabel(t *testing.T) {
 	r := require.New(t)
 	stringHash := func(data []byte) []byte {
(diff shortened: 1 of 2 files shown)


shared/shared_test.go
```diff
@@ -53,6 +53,12 @@ func TestFiatShamirReturnsCorrectNumOfIndices(t *testing.T) {
 	require.Len(t, FiatShamir(challenge, 20, 15), 15)
 }
 
+// This test used to hang.
+// See https://github.com/spacemeshos/poet/issues/173
```
fasmat, line 57: I think this can be removed.

fasmat (approved):

fasmat:
This solves the issue that the function takes very long if `spaceSize` is only marginally larger than `securityParam`.

While I'm ok with the change itself, I don't think this is an issue: `spaceSize` (on testnets 2048 - 8192, mainnet will be at least 275*10^9) is several magnitudes larger than `securityParam` (at most 255).

shared/shared_test.go
poszu: Why? The comment describes the purpose of this test (which is to test a known past issue).

Issue #185, opened 2022-12-12
`TestService_Recovery` is flaky


poszu:
Test sporadically fails on Mac and Windows.

```
=== FAIL: service TestService_Recovery (2.89s)
2022-12-12T08:29:30.412Z	INFO	Spacemesh.recovery	Recovering service state	{"datadir": "/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/TestSubmitIdempotency3380714174/001"}
2022-12-12T08:29:30.424Z	INFO	Spacemesh	creating poet service. min memory layer: 17. genesis: 2022-12-12T08:29:31Z
2022-12-12T08:29:30.426Z	INFO	Spacemesh	Service public key: eaad52887898fb6d20e66e0ac79a1f86eb2908c7ad83cb7868f1ddfe6b08fc76
2022-12-12T08:29:30.426Z	INFO	Spacemesh.recovery	Recovering service state	{"datadir": "/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/TestService_Recovery2546722483/001"}
2022-12-12T08:29:30.508Z	INFO	Spacemesh	Round opened	{"ID": "0"}
2022-12-12T08:29:30.508Z	INFO	Spacemesh.worker	waiting for execution to start	{"wait time": "991.281ms", "round": "0"}
2022-12-12T08:29:30.599Z	INFO	Spacemesh	Round opened	{"ID": "0"}
2022-12-12T08:29:30.599Z	INFO	Spacemesh.worker	waiting for execution to start	{"wait time": "900.474ms", "round": "0"}
2022-12-12T08:29:30.614Z	INFO	Spacemesh	service shutting down
2022-12-12T08:29:30.614Z	INFO	Spacemesh.worker	service shutting down
2022-12-12T08:29:30.615Z	INFO	Spacemesh	Round 0 torn down (cleanup false)
2022-12-12T08:29:30.615Z	INFO	Spacemesh	service shutdown complete
    service_test.go:130: 
        	Error Trace:	/Users/runner/work/poet/poet/service/service_test.go:130
        	Error:      	Not equal: 
        	            	expected: "0"
        	            	actual  : "1"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-0
        	            	+1
        	Test:       	TestService_Recovery
```

poszu:
Closed in #196

PR #189, opened 2022-12-14
Update spacemesh API

fasmat:
Update poet with the latest spacemesh API


fasmat:
@poszu the unit tests on windows are very unstable. I executed them 4 times and got different errors every time:

- db file is used by another process
- file seeking fails due to EOF
- when asserting the round ID it is already 2 when it should be 1
- after recovery there are no executing rounds when there should be one

I reverted the last merge in this PR, so that we can get a green build again.

poszu:
> @poszu the unit tests on windows are very unstable. I executed them 4 times and got different errors every time:

I am aware of these issues with Mac and Windows. Unfortunately, the tests were written with some timing assumptions in mind and:
- the mac runner is slow,
- time on windows has low res.

> * db file is used by another process

This is caused by asynchronous round teardown. I plan to fix it in #181.

> * file seeking fails due to EOF

This is caused by generating empty proofs (0 leaves) - it happens on Mac as it is slow and the round is ended the moment it starts. I don't have a clear idea of how to fix it. Perhaps an empty proof is a valid one and it should be broadcasted empty instead of trying to build leaves and nodes?

> * when asserting the round ID it is already 2 when it should be 1
> * after recovery there are no executing rounds when there should be one

It's tracked in #185. 

> I reverted the last merge in this PR, so that we can get a green build again.

These issues are not directly related to the PR you revert. Reverting doesn't fix the problem, it will persist, perhaps a little less often though. I'm against reverting it.


I'd rather stop running tests on mac and windows as they are not the platform that Poets will run on. There's no point in supporting them.

poszu:
Why is it needed to update `api` repo now? Poet only uses the Gateway Service and it's unchanged between v1.5.2 and v1.5.3.

fasmat:
> These issues are not directly related to the PR you revert. Reverting doesn't fix the problem, it will persist, perhaps a little less often though. I'm against reverting it.

That might very well be the case but at the moment `develop` has a red build, and the PR I reverted caused it to become red: https://github.com/spacemeshos/poet/actions/runs/3684903021
This blocks other PRs like this one, that's why I reverted it.

> I'd rather stop running tests on mac and windows as they are not the platform that Poets will run on. There's no point in supporting them.

I'm against this. Poet is a dependency (as a library not just as a service) of go-spacemesh. go-spacemesh needs to support all 3 platforms, so poet should also be tested for all three. We can maybe skip some of the tests on mac and windows if they are only needed for the service but not for the functionality that go-spacemesh uses.

I'd rather stop running tests in parallel if this is unstable until the issues listed above (which seem to be already tracked) are resolved. After that they can be enabled again. The PR that introduced the parallel tests was just unlucky to have a single green build, you can try re-running the CI on the PR and it will fail more often then pass - imo the PR should not have been merged at the current state.

> Why is it needed to update api repo now? Poet only uses the Gateway Service and it's unchanged between v1.5.2 and v1.5.3.

I initially thought it is required to update the API to `v1.5.3` because I had poet related issues in the systests related to a PR I'm trying to merge, but the issue isn't Poet; it's unstable tests. Updating the API isn't required at the moment, but it also shouldn't hurt to do so.

poszu:
>>These issues are not directly related to the PR you revert. Reverting doesn't fix the problem, it will persist, perhaps a little less often though. I'm against reverting it.
>
>That might very well be the case but at the moment develop has a red build, and the PR I reverted caused it to become red: https://github.com/spacemeshos/poet/actions/runs/3684903021
This blocks other PRs like this one, that's why I reverted it.

Again, the failures are not caused by running tests in parallel. Tests are flaky nevertheless and the build will continue to fail even after it's reverted. Running tests in parallel is not unstable. The tests are unstable in general.


>> I'd rather stop running tests on mac and windows as they are not the platform that Poets will run on. There's no point in supporting them.
> 
> I'm against this. Poet is a dependency (as a library not just as a service) of go-spacemesh. go-spacemesh needs to support all 3 platforms, so poet should also be tested for all three. We can maybe skip some of the tests on mac and windows if they are only needed for the service but not for the functionality that go-spacemesh uses.

IMHO, the Poet should not be a dependency of go-spacemesh as a whole or even at all. It's used in two places only:
- unit tests use the integration harness. The poet could (and I think it should) be instead run in a docker container. Alternatively, integration tests could be left for systests.
- to verify proof validity using `verifier.Validate()`. I think that this could be inlined in the go-spacemesh repo. If we were to run tests on mac and win selectively, then only these tests need to be run: poet/verifier/verifier_test.go.

fasmat:
> Again, the failures are not caused by running tests in parallel. Tests are flaky nevertheless and the build will continue to fail even after it's reverted. Running tests in parallel is not unstable. The tests are unstable in general.

Yes, but running them in parallel causes them more often to fail. The build is red, reverting the PR makes the build green. Shall we now not merge any PRs until the issues are resolved?

> IMHO, the Poet should not be a dependency of go-spacemesh as a whole or even at all. It's used in two places only:

This is not the case yet; poet is used on all 3 platforms at the moment. Removing the tests is better than leaving them and just making it less likely that they randomly fail?

poszu:
> The build is red, reverting the PR makes the build green.

IMO it was luck.

poszu (approved):

PR #192, opened 2022-12-15
Make round teardown synchronous

poszu:
Part of #181

### DIFF
--- a/service/round_test.go
+++ b/service/round_test.go
@@ -44,7 +44,7 @@ func TestRound_Recovery(t *testing.T) {
 	req.NoError(err)
 
 	// Execute r1 as a reference round.
-	r1, err := newRound(ctx, tmpdir, 0)
+	r1, err := newRound(tmpdir, 0)
 	req.NoError(err)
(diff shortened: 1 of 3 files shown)


service/round_test.go
```diff
@@ -87,10 +88,11 @@ func TestRound_Recovery(t *testing.T) {
 
 	stop()
 	req.ErrorIs(r2recovery1.recoverExecution(ctx, state.Execution, time.Now().Add(duration)), prover.ErrShutdownRequested)
-	require.NoError(t, r2recovery1.waitTeardown(context.Background()))
+	req.NoError(r2recovery1.teardown(false))
 
 	// Recover r2 execution again, and let it complete.
 	ctx, stop = context.WithCancel(context.Background())
+	defer stop()
```
fasmat, line 96: 👍 there are a few more `context.WithCancel(context.Background())` calls in this test. Can we add a `defer stop()` to those too? (In case a test is terminated early due to a failing assert?)

fasmat (approved):
Just a minor suggestion, lgtm

PR #195, opened 2022-12-19
Refactors in `round.go`

### LINKED ISSUE #181: Refactor `round`

poszu:
Closes #181 

- Simplified `round struct` a lot:
  * removed unnecessary channels,
  * simplified state, removed duplication,
  * found and fixed a few bugs in state persistence
- Added unit tests and cleaned the existing ones for better readability


service/service.go
```diff
@@ -277,7 +271,8 @@ func (s *Service) Run(ctx context.Context) error {
 		}
 	}
 
-	return s.loop(ctx, toResume)
+	err := s.loop(ctx, toResume)
+	return err
```
poszu:
```suggestion
	return s.loop(ctx, toResume)
```

service/round.go
```diff
@@ -121,65 +96,44 @@ func (r *round) submit(key, challenge []byte) error {
...
-	defer iter.Release()
-	return !iter.Next()
-}
-
 func (r *round) execute(ctx context.Context, end time.Time, minMemoryLayer uint) error {
 	logger := logging.FromContext(ctx).WithFields(log.String("round", r.ID))
 	logger.Info("executing until %v...", end)
 
-	r.executionStarted = time.Now()
+	r.ExecutionStarted = time.Now()
```
fasmat: Should `executionStarted` be private? Otherwise anyone with access to round can modify that field.

service/round.go
```diff
@@ -121,65 +96,44 @@ func (r *round) submit(key, challenge []byte) error {
...
+	numLeaves, nip, err := prover.GenerateProof(
 		ctx,
 		r.datadir,
-		hash.GenLabelHashFunc(r.execution.Statement),
-		hash.GenMerkleHashFunc(r.execution.Statement),
+		hash.GenLabelHashFunc(r.Execution.Statement),
+		hash.GenMerkleHashFunc(r.Execution.Statement),
 		end,
-		r.execution.SecurityParam,
+		r.Execution.SecurityParam,
```
fasmat: Same here, `Execution` can be modified from outside of `round`.

service/round.go
```diff
@@ -191,108 +145,75 @@ func (r *round) persistExecution(tree *merkle.Tree, treeCache *cache.Writer, num
...
+	if err := load(filename, &state); err != nil {
+		return err
 	}
-	if r.execution.SecurityParam != s.Execution.SecurityParam {
-		return nil, errors.New("SecurityParam config mismatch")
+	if r.Execution.SecurityParam != state.Execution.SecurityParam {
+		return errors.New("SecurityParam config mismatch")
 	}
-	r.stateCache = s
+	r.roundState = state
```
fasmat: Suggestion: the type `roundState` is only used for (de-)serialization of the state to/from disk but is not a field of the round.

Instead the round has `ExecutionStarted` and `Execution`  as private fields which are set here.

service/round.go
```diff
@@ -191,108 +145,75 @@ func (r *round) persistExecution(tree *merkle.Tree, treeCache *cache.Writer, num
...
 
 func (r *round) saveState() error {
 	filename := filepath.Join(r.datadir, roundStateFileBaseName)
-	v := &roundState{
-		Opened:           r.opened,
-		ExecutionStarted: r.executionStarted,
-		Execution:        r.execution,
-	}
-	return persist(filename, v)
+	return persist(filename, &r.roundState)
```
fasmat: same here, I'd prefer the way it was before: create a `roundState` object that is used to persist the state, but is not part of the round itself.

service/round_test.go
```diff
@@ -2,227 +2,322 @@ package service
...
-		req.NoError(r1.submit(ch, ch))
-	}
-	req.Equal(len(challenges), r1.numChallenges())
-	req.False(r1.isEmpty())
+func newTestRound(t *testing.T) *round {
+	t.Helper()
+	round, err := newRound(t.TempDir(), rand.Uint32())
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		require.NoError(t, round.teardown(true))
```
fasmat: `assert` instead of `require` otherwise the `t.Cleanup` will stop executing possible other cleanup steps if this one fails.

service/round_test.go
```diff
@@ -2,227 +2,322 @@ package service
...
-	ctx, stop = context.WithCancel(context.Background())
-	defer stop()
-	r2recovery1, err := newRound(tmpdir, 1)
+	// Act
+	round, err := newRound(t.TempDir(), 7)
 	req.NoError(err)
-	req.Equal(len(challenges), r2recovery1.numChallenges())
-	req.False(r2recovery1.isEmpty())
+	t.Cleanup(func() {
+		req.NoError(round.teardown(true))
```
fasmat: same here: `assert` over `require` in cleanup steps.

service/round_test.go
```diff
@@ -2,227 +2,322 @@ package service
...
-	req.NoError(err)
+		// Act
+		for _, ch := range challenges {
+			req.NoError(round.submit(ch, ch))
+		}
 
-	req.NoError(r2recovery2.recoverExecution(ctx, state.Execution, time.Now().Add(duration)))
-	req.NoError(r2recovery2.teardown(true))
+		// Verify
+		req.EqualValues(len(challenges), numChallenges(round))
```
fasmat: I think both are `int` and should also pass `req.Equal()`.

This should be the case for most uses of `EqualValues` in this file.

service/round_test.go
```diff
@@ -2,227 +2,322 @@ package service
...
+
+	req.NoError(
+		verifier.Validate(
+			*execution.NIP,
+			hash.GenLabelHashFunc(execution.Statement),
+			hash.GenMerkleHashFunc(execution.Statement),
+			execution.NumLeaves,
+			shared.T,
+		),
+	)
```
fasmat: Can this be shortened to just the last assertion? I assume `verifier.Validate` returns an error if any of the previous assertions would fail.

service/round_test.go
```diff
@@ -2,227 +2,322 @@ package service
...
+		req.NoError(round.submit([]byte("key"), challenge))
+		ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*100)
+		defer cancel()
+		req.ErrorIs(round.execute(ctx, time.Now().Add(time.Hour), 1), prover.ErrShutdownRequested)
+		req.NoError(round.teardown(false))
+
+		// Act
+		recovered, err := newRound(tmpdir, 0)
+		req.NoError(err)
+		t.Cleanup(func() { req.NoError(recovered.teardown(false)) })
```
fasmat: `assert` over `require`.

service/round_test.go
```diff
@@ -2,227 +2,322 @@ package service
...
+
+		// Arrange
+		round, err := newRound(tmpdir, 0)
+		req.NoError(err)
+		req.NoError(round.teardown(false))
+
+		// Act
+		recovered, err := newRound(tmpdir, 0)
+		req.NoError(err)
+		t.Cleanup(func() { req.NoError(recovered.teardown(false)) })
```
fasmat: `assert` over `require`.

service/round.go
poszu: I'd like it to be private too, but then serialization doesn't work (which is required to persist round's state). Do you know how to get around it?

service/round.go
poszu: It could be modified outside of `round` either way as Go has no notion of "private vs public", just "unexported vs exported on a package level". `round struct` is unexported so it's only available inside the `service` pkg.

service/round.go
fasmat: See [here](https://github.com/spacemeshos/poet/pull/195#discussion_r1054263010) and [here](https://github.com/spacemeshos/poet/pull/195#discussion_r1054263737) 😉

service/round.go
poszu: I know. It was a conscious choice to put the _persistable_ state together in a struct. This way it's immediately clear what the _persistable_ state is. It's embedded anyway so there is no _visual_ cost of having to go through an extra level of nesting to work with the data.

fasmat: I can't follow your argument? What does the layout of the json/xdr file used for persistence have to do with the structure of the round object? Why does this have to be the same?

No part of the code besides `loadState()` and `saveState()` need to know what is persisted and how.

poszu: The layout of xdr file has nothing to do with it. The argument is that **it is clear what data is persisted**.

fasmat: That's clear without having a subtype inside the round imo, also why does this need to be visible outside of the two methods I stated?

poszu: It's less clear to me. Why is this such an issue?

fasmat: I don't see the benefit of having `roundstate` as an object. You say so it's clear what is persisted, but it comes at the drawback that anyone with access to round can modify its `ExecutionState` directly, which was not possible before.

`loadState()` and `saveState()` were very straight forward before and showed which data was persisted and how; this is now hidden away in `roundstate` with the additional drawback stated above. I don't see the benefit of this change.

poszu: I see a benefit as explained.

Regarding:
> it comes at the drawback that anyone with access to round can modify its ExecutionState directly, which was not possible before.

It is not true. It has always been possible to modify the state from the outside as there is no concept of "private" state in Golang whatsoever. Everybody from `package service` can and always could modify the state of `round struct`. And as it is unexported, no one from outside of `package service` can hold a `round`. Please correct me if I'm wrong and why.

fasmat: Yes it is true. ExecutionState can be modified by anyone. It is public now and wasn't before. I am aware that before everyone inside the `package service` could modify it, but now everyone in any package can modify this value. And `round` doesn't need to be exported to be used outside of the package. I sent you an exact example.

There is no need to DRY up this code. I prefer the code how it was before and it didn't have the drawback of exposing the state publicly.

service/round.go
poszu: Changed. PTAL

service/round.go
poszu: changed.

poszu:
@fasmat I changed accordingly to your comments. Please, take a look again.

fasmat (approved):
❤️  lgtm

PR #207, opened 2022-12-30
Run gosec in CI

### LINKED ISSUE #206: Run gosec in CI
We should run https://github.com/securego/gosec in CI to detect security vulnerabilities automatically.

poszu:


integration/exec.go
```diff
@@ -39,9 +39,7 @@ func poetExecutablePath(baseDir string) (string, error) {
 		outputPath += ".exe"
 	}
 
-	cmd := exec.Command(
-		"go", "build", "-o", outputPath, "github.com/spacemeshos/poet",
-	)
+	cmd := exec.Command("go", "build", "-o", outputPath, "github.com/spacemeshos/poet") //#nosec G204
```
fasmat, line 42: can we have a proper fix of the issue here? `outputPath` is `basePath` + "/poet", but `basePath` is already `/tmp/poet`.

`baseDir()` is only called in a single place to have its return value then passed into this function. imo `baseDir()` can just be inlined here instead of passing `baseDir` into this function, something like:

```go
outputPath := filepath.Join(os.TempDir(), "poet")
if err := os.MkdirAll(baseDir, 0o755); err != nil {
    return err
}

cmd := exec.Command("go", "build", "-o", outputPath, "github.com/spacemeshos/poet")
```

should get rid of the complaint by gosec and also makes the code less spread out into multiple places

cmd/bench/bench.go
```diff
@@ -36,7 +36,7 @@ func main() {
 		profFilePath := path.Join(dir, "./CPU.prof")
 		fmt.Printf("CPU profile: %s\n", profFilePath)
 
-		f, err := os.Create(profFilePath)
+		f, err := os.Create(profFilePath) //#nosec G304
```
fasmat: this can just be

```suggestion
		f, err := os.Create("./CPU.prof")
```

fasmat (commented):
I'm not sure about enabling `gosec` and then ignoring most issues it finds with comments in the code. What's the benefit of doing the analysis then?

integration/exec.go
poszu: The `G204` is about executing `exec.Command` with a variable. It literally says:
>G204: Subprocess launched with variable

The change you propose doesn't fix it. G204 is intended for code audits and can be disabled after the reported code is checked. Please see the discussion in https://github.com/securego/gosec/issues/292.

cmd/bench/bench.go
poszu: Great, thanks! I slightly modified it for logging purposes.

poszu:
> I'm not sure about enabling `gosec` and then ignoring most issues it finds with comments in the code. What's the benefit of doing the analysis then?

I'm not ignoring most issues. The `gosec` caught some significant problems and they were fixed. The ignored ones are `G204` and `G304` which complain about using variables in process execution or opening files. These were audited. They aren't an issue for us because:
- they are in code for integration tests,
- we don't use data passed from users.

IMO it is a huge gain to have `gosec` running in CI.

integration/exec.go
fasmat: Not happy how gosec handles `G204`. This now means if we introduce a potential error where outputPath can be compromised it won't even warn us about it any more due to the suppressed warning.

But since this is the intended way how to handle it, I guess we have to leave it like this.

fasmat:
I'm not arguing against the checks themselves, I just don't like that the only way to "fix" some found errors is to disable the check for them where they are found. If the code around the line where the issue was found changes e.g. to use data passed from users `gosec` won't warn us again that there might be an issue.

But since this is more an issue with `gosec` than with the changes in this PR, I approve 🙂

fasmat (approved):

PR #209, opened 2023-01-03
Lock round execution to OS thread and write TID to file

poszu:
This change locks the OS thread of the goroutine executing round (generating the PoET). It prevents the scheduler from moving the goroutine to a different thread, which improves performance by stabilizing the CPU core boost.

The thread ID of the locked goroutine is written into a file so that an operator can pick it up and lock the OS thread to a CPU core to prevent it from jumping from core to core.


service/tid/tid_darwin.go
```diff
@@ -0,0 +1,8 @@
+//go:build darwin
+// +build darwin
```
fasmat, line 2: `// +build darwin` (and other OS) was the annotation style pre-1.17 and is ignored by go starting with Go 1.18. I think we can remove these lines.

fasmat (commented):
The locking and unlocking looks fine to me, but why are we writing the TID to a file? Is this information used somewhere?

poszu:
> The locking and unlocking looks fine to me, but why are we writing the TID to a file? Is this information used somewhere?

Sorry, I forgot to explain it. Added explanation in PR's description.

fasmat (approved):

PR #215, opened 2023-01-12
Avoid calling time.Until in the POSW loop

poszu:
It improves leaves per second by around 2%.

### DIFF
--- a/prover/prover.go
+++ b/prover/prover.go
@@ -2,7 +2,6 @@ package prover
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -30,8 +29,6 @@ const (
 	// to allow potential crash recovery.
 	hardShutdownCheckpointRate = 1 << 24
 )
-
-var ErrShutdownRequested = errors.New("shutdown requested")
 
 type persistFunc func(ctx context.Context, tree *merkle.Tree, treeCache *cache.Writer, nextLeafId uint64) error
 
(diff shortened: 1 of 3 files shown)


prover/prover.go
```diff
@@ -191,47 +192,74 @@ func makeRecoveryProofTree(
...
+	makeLabel := shared.MakeLabelFunc()
+	for leafID := nextLeafID; ; leafID++ {
 		select {
 		case <-ctx.Done():
 			if err := persist(ctx, tree, treeCache, leafID); err != nil {
-				return 0, nil, fmt.Errorf("%w: error happened during persisting: %v", ErrShutdownRequested, err)
+				return 0, fmt.Errorf("persisting execution state: %w", err)
 			}
-			return 0, nil, ErrShutdownRequested
+			return leaves, nil
```
fasmat: imo when returning early due to `ctx.Done()` the returned error should be `ctx.Err()`

prover/prover.go
```diff
@@ -191,47 +192,74 @@ func makeRecoveryProofTree(
...
+	defer cancel()
+	leaves, err := sequentialWork(poswCtx, labelHashFunc, tree, treeCache, nextLeafID, securityParam, persist)
+	select {
+	case <-ctx.Done():
+		if err != nil {
+			return 0, nil, fmt.Errorf("%w: %v", ErrShutdownRequested, err)
+		}
+		return 0, nil, ErrShutdownRequested
+	default:
+	}
```
fasmat: I would move that into the `sequentialWork` function; on line 212:

```golang
return leaves, ctx.Err()
```

A `context.Cancelled` error already implies that a shutdown / cancellation was requested. The dedicated type `ErrShoutdownRequested` we currently have for it could just replaced in tests with `context.Cancelled`.

fasmat (approved):
Some small suggestions on error propagation.

prover/prover.go
poszu: The thing is that this condition is not "returning early". This is the main return point meaning "the work was done". When the `ctx` in `sequentialWork` is cancelled it means "stop producing more leaves". It was achieved via `time.Until(end) <= 0` before. Now the `ctx` is cancelled instead.

prover/prover.go
poszu: Removed `ErrShoutdownRequested`

fasmat (approved):
LGTM

PR #223, opened 2023-01-18
Bump merkle-tree - v0.2.0

poszu:
Before:
```
BenchmarkGetProof-20    	       1	30006684789 ns/op	   5.30 MB/s	   4965833 leafs/op	    165528 leafs/sec	10519326320 B/op	44632446 allocs/op
```

After:
```
BenchmarkGetProof-20    	       1	30029381371 ns/op	   5.54 MB/s	   5195681 leafs/op	    173189 leafs/sec	977521384 B/op	    6326 allocs/op
```

### DIFF
--- a/hash/hash.go
+++ b/hash/hash.go
@@ -2,6 +2,7 @@ package hash
 
 import (
 	"github.com/minio/sha256-simd"
+	"github.com/spacemeshos/merkle-tree"
 )
 
 // LabelHashNestingDepth is the number of recursive hashes per label.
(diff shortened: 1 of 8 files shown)


pigmej (approved):

hash/hash.go
```diff
@@ -12,32 +13,33 @@ const LabelHashNestingDepth = 100
...
 	return func(data []byte) []byte {
-		message := append(challenge, data...)
-		var res [32]byte
-		for i := 0; i < LabelHashNestingDepth; i++ {
-			res = sha256.Sum256(message)
-			message = res[:]
+		h.Reset()
+		_, _ = h.Write(challenge)
+		_, _ = h.Write(data)
+		hash := h.Sum(hashBuf[:0])
```
fasmat, line 36: would `h.Sum(nil)` here not have the same effect?

hash/hash.go
```diff
@@ -12,32 +13,33 @@ const LabelHashNestingDepth = 100
...
+
+		for i := 1; i < LabelHashNestingDepth; i++ {
+			h.Reset()
+			_, _ = h.Write(hash)
+			hash = h.Sum(hash[:0])
```
fasmat, line 41: same as above; `h.Sum(nil)` should have the same effect.

prover/prover.go
```diff
@@ -232,7 +233,8 @@ func sequentialWork(
 
 	for {
 		// Generate the next leaf.
-		err := tree.AddLeaf(makeLabel(labelHashFunc, nextLeafID, tree.GetParkedNodes()))
+		parkedNodes = tree.GetParkedNodes(parkedNodes[:0])
```
fasmat, line 236: again, I'm no 100% sure about this but `parkedNodes[:0]` is a slice with (in this case) len 0 and cap 0, `nil` should have the same effect.

service/round.go
```diff
@@ -149,7 +149,7 @@ func (r *round) persistExecution(
 	}
 
 	r.execution.NumLeaves = numLeaves
-	r.execution.ParkedNodes = tree.GetParkedNodes()
+	r.execution.ParkedNodes = tree.GetParkedNodes(r.execution.ParkedNodes[:0])
```
fasmat, line 152: same as my previous comment.

prover/prover.go
poszu: A `parkedNodes[:0]` is a slice with length 0 but capacity equal to capacity of `parkedNodes`. The idea is that `parkedNodes` local is reused in each loop iteration avoiding reallocating memory every time (the result is reassigned back to `parkedNodes` exactly like `append()` works.

hash/hash.go
poszu: No, passing a `nil` would cause `Sum()` to allocate memory. The idea is to reuse the memory in `hashBuf` array. The `hash` slice will point to `hashBuf`.

poszu: The difference is significant:
with `nil`:

> BenchmarkGetProof-20    	       1	30001720023 ns/op	   4.49 MB/s	   4205877 leafs/op	    140196 leafs/sec	14240215608 B/op	420594083 allocs/op

with reusing the buffer:
> BenchmarkGetProof-20    	       1	30001170355 ns/op	   5.57 MB/s	   5218087 leafs/op	    173936 leafs/sec	977524200 B/op	    6333 allocs/op

fasmat: `hashBuf` is an array that gets used as the backing array of a new slice here. It is true that calling the returned closure multiple times saves allocating memory, but I think it would be more efficient for the runtime if `hashBuf` would be a slice instead of an array: `hashBuf := make([]byte, 0, sha256.Size)` not requiring the "cast" from array to slice on every call here.

prover/prover.go
fasmat: I see, can the capacity be estimated before calling this function? On the first call `parkedNodes` is `nil` and the slice is created here by essentially `nil[:0]` with `cap` 0 and `len` 0 (to then be re-allocated by the `tree.GetParkedNodes` function).

If `tree.GetParkedNodes` allocates enough capacity this will only happen here and is only 1 allocation that could be saved on the first iteration of the loop.

hash/hash.go
fasmat: correction: this seems to be correct.

prover/prover.go
poszu: Yes, you are right - it's `nil` on the first iteration. Then it is resized (re-allocated) if required in `tree.GetParkedNodes()`. It is, however, difficult to predict the required size for the slice. It grows with the tree but rather slowly. IIRC, it's 1 parked node per tree layer which is log2(tree width) - negligible (according to my quick calculation it will be max 38 elements for 300k leaves/sec and 2w epoch).

PR #224, opened 2023-01-19
Improve hashing speed

poszu:
## Changes
- bumped sha256-simd to our fork
- rewritten the tree hashing functions to use our fork
- refactored FiatShamir to better use sha256 to improve speed.


hash/hash.go
```diff
@@ -12,34 +14,66 @@ const LabelHashNestingDepth = 100
...
+			h.CheckSumInto((*[sha256.Size]byte)(buf))
+			return buf
+		}
+	default:
+		return func(buf, lChild, rChild []byte) []byte {
+			h.Reset()
+			_, _ = h.Write(challenge)
+			_, _ = h.Write(lChild)
+			_, _ = h.Write(rChild)
+			return h.Sum(buf)
```
fasmat: I don't think this is correct, this should probably be

```suggestion
			return h.Sum(buf[:0])
```

otherwise the content of `buf` will be used as input to the hash.

hash/hash_test.go
```diff
@@ -37,3 +37,31 @@ func TestGenMerkleHashFunc(t *testing.T) {
...
+
+	require.Equal(
+		t,
+		[]byte{
+			104, 1, 230, 180, 3, 202, 139, 10, 164, 69, 74, 169, 116, 77, 27, 134,
+			138, 163, 204, 32, 182, 197, 205, 194, 13, 48, 30, 224, 254, 1, 46, 133,
+		},
+		GenLabelHashFunc(challenge)(data),
+	)
+}
```
fasmat: Depending on the system these tests are executed they only test one of the two implementations.

You could extract the 2 cases of the `switch` in `GenLabelHashFunc` as private methods, call them explicitly in one test and compare if their output is the same. Similarly for `GenMerkleHashFunc` below.

poszu: Yes, but anyway it's impossible to test the optimized version if host doesn't support it. It would:
- test both if host supports it
- test only generic if host doesn't support it.

fasmat: The important part imo that we have a test that compares the output of both implementations and makes sure they are the same even if that test is only executed on platforms that support the optimized version.

poszu: Done

fasmat (approved):
LGTM

PR #225, opened 2023-01-19
Increase leaves counter metric every leaf.

poszu:
Note: benchmarks doesn't show any difference in speed.


pigmej (approved):
Let's do the micro opts later ;)

PR #226, opened 2023-01-19
Removed tools from source & dependency graph

fasmat:
Using a `tools.go` file was the recommended approach to tracking versions of tools before `go install` became version aware.

The problem with using `tools.go` and tracking versions of dependencies in the `go.mod` file is that the dependency graphs of poet and the tools merge. This means adding version `1.2.0` of `google.golang.org/grpc/cmd/protoc-gen-go-grpc` to the dependencies will not cause version `1.2.0` of the tool to be installed. It will only use the source of 1.2.0 and the dependency graph of poet to install it (causing a hybrid version that is different from the one that would be installed using `go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2.0`


Makefile
```diff
@@ -79,11 +79,10 @@ endif
...
 # Download protoc plugins
 protoc-plugins:
-	@go install \
-		github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway \
-		github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2 \
-		google.golang.org/protobuf/cmd/protoc-gen-go \
-		google.golang.org/grpc/cmd/protoc-gen-go-grpc
+	@go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway@v2.15.0
+	@go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2@v2.15.0
+	@go install google.golang.org/protobuf/cmd/protoc-gen-go@b0a9446
```
fasmat, line 84: `progoc-gen-go` isn't correctly tagged in `google.golang.org/protobuf`: the tag `cmd/protoc-gen-go/v1.28.1` is missing, `b0a9446` is the commit tagged with that version.

Makefile
```diff
@@ -52,10 +52,10 @@ GOVULNCHECK := $(GOBIN)/govulncheck
 GOLINES := $(GOBIN)/golines
 
 $(GOVULNCHECK):
-	@go install golang.org/x/vuln/cmd/govulncheck
+	@go install golang.org/x/vuln/cmd/govulncheck@latest
```
fasmat, line 55: `govulncheck` has no official released version yet.

poszu:
It is a bummer that the tools bloat dependencies of the package itself, I noticed it too :(

On the other hand with the _tools approach_, all versions are tracked in a single place (go.mod) compared to Makefile and go.mod after this change. That's why I think the _tools one_ is better after all.

fasmat:
The problem is you are not installing `v1.2.0` as released by the author of the tool, you are just using the source of `v1.2.0` and overwriting the dependencies with the ones used by poet if you track the version with `tools.go` / `go.mod`.

EDIT: this can in the worst case lead to the installation of the tool randomly failing in the future or preventing us from updating dependencies of poet.

go.mod
```diff
@@ -12,65 +12,46 @@ require (
...
 	github.com/stretchr/testify v1.8.1
 	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	go.uber.org/zap v1.24.0
-	golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e
+	golang.org/x/exp v0.0.0-20230116083435-1de6713980de
 	golang.org/x/sync v0.1.0
-	golang.org/x/sys v0.4.0
-	golang.org/x/vuln v0.0.0-20221222221150-61d83dad62c1
-	google.golang.org/genproto v0.0.0-20221207170731-23e4bf6bdc37
+	golang.org/x/sys v0.3.0
```
poszu: Why is it downgraded?

fasmat: Not intentionally, fixed.

poszu (approved):

PR #227, opened 2023-01-19
Configurable tree file-based layers buffer size

poszu:
Added two new configuration flags:
1. `--lps` to configure the estimated leaves per second (it's different per platform)
It's important for the poet service to estimate the size of proof tree upfront.
2. `--tree-file-buffer` to configure the buffer size for file-base (bottom) layers of the proof tree.


cmd/bench/bench.go
```diff
@@ -46,12 +46,14 @@ func main() {
 	end := proofGenStarted.Add(cfg.Duration)
 	leafs, merkleProof, err := prover.GenerateProofWithoutPersistency(
 		context.Background(),
-		tempdir,
+		prover.TreeConfig{
+			Datadir:           tempdir,
+			FileWriterBufSize: 0,
```
fasmat: NIT: 0 is the default value for `uint` so this line could be omitted.

prover/prover.go
```diff
@@ -81,22 +86,20 @@ func GenerateProofRecovery(
...
-func makeProofTree(
-	datadir string,
-	merkleHashFunc merkle.HashFunc,
-	minMemoryLayer uint,
-) (*merkle.Tree, *cache.Writer, error) {
-	metaFactory := NewReadWriterMetaFactory(minMemoryLayer, datadir)
+func makeProofTree(treeCfg TreeConfig, merkleHashFunc merkle.HashFunc) (*merkle.Tree, *cache.Writer, error) {
+	if treeCfg.MinMemoryLayer < LowestMerkleMinMemoryLayer {
+		treeCfg.MinMemoryLayer = LowestMerkleMinMemoryLayer
+	}
```
fasmat, line 101: Is this behaviour documented somewhere?

fasmat (approved):

pigmej (approved):

prover/prover.go
poszu: I don't think it is. It just means that at least 1 tree layer (the base layer) will be stored on a disk. Please note, that I just moved this condition from `service.go` (there is no behaviour change).

PR #236, opened 2023-01-27
Add GRPC API for querying the list of members for a given round

### LINKED ISSUE #200: Add RPC API for querying the list of members for a given round.
This info should be available upon the completion of the round's registration, and the beginning of its actual execution.

poszu:
Closes #200.

- the list of members is available after a round started executing,
- the proof obtained via `GetProof()` GRPC doesn't contain a list of members anymore.


service/db.go
```diff
@@ -14,52 +14,93 @@ import (
...
 
-type ProofsDatabase struct {
+type ServiceDatabase struct {
 	db     *leveldb.DB
 	proofs <-chan shared.ProofMessage
+	rounds <-chan RoundData
+}
+
+func makeKey(t dataType, k []byte) []byte {
+	key := make([]byte, len(k)+1)
```
fasmat, line 33: This should probably be

```suggestion
	key := make([]byte, 0, len(k)+1)
```

otherwise the returned `key` is `2*len(k) + 2` bytes long where the first `len(k)+1` bytes are all `0x00`

service/db.go
```diff
@@ -14,52 +14,93 @@ import (
...
-		return nil, fmt.Errorf("failed to open database @ %s: %w", dbPath, err)
+		return nil, fmt.Errorf("opening database @ %s: %w", dbPath, err)
 	}
 
-	return &ProofsDatabase{db, proofs}, nil
+	return &ServiceDatabase{db, proofs, rounds}, nil
 }
 
-func (db *ProofsDatabase) Run(ctx context.Context) error {
+func (db *ServiceDatabase) Run(ctx context.Context) error {
```
fasmat, line 77: I feel like the service does more than it should. A database persists and fetches the entities from the store, but imo shouldn't have a `Run` method that kicks off an endless loop for processing data.

Suggestion: instead of having this loop here (that needs to be executed on its own go-routine), how about if the writers to the channels `db.proofs` and `db.rounds` just call `AddProof` and `AddRound` on the `ServiceDB`? Then no endless loop running on its own routine synchronising between channels is needed.

LevelDB is according to its documentation even save to be used concurrently, so the whole synchronization to avoid calling `Put` concurrently isn't even necessary (and if it would I would prefer just using a mutex in `ServiceDatabase`).

This would also allow to simplify the tests in `db_test.go`.

service/service.go
```diff
@@ -232,6 +248,16 @@ func (s *Service) loop(ctx context.Context, roundToResume *round) error {
...
 			end := s.roundEndTime(round)
+
+			select {
+			case s.rounds <- RoundData{
+				ID:      round.ID,
+				Members: round.Members(),
+			}:
+			default:
+				logger.Debug("could not push round data, the receiving end might've hang")
+			}
```
fasmat, line 259: imo this should just be `servicedatabase.AddRound(round.ID, round.Members())`.

Should persisting a round fail if the DB is at the same moment persisting a proof or another round?

fasmat (commented):
I think we should discuss the design of the `ServiceDB` and if we really need the synchronization that is currently built into it. I feel like this is not required since `levelDB` is already safe to be used concurrently.

What do you think?

service/db.go
poszu: Yes, thanks!

poszu:
> I think we should discuss the design of the `ServiceDB` and if we really need the synchronization that is currently built into it. I feel like this is not required since `levelDB` is already safe to be used concurrently.
> 
> What do you think?

Synchronization for safe concurrent use is absolutely not the point of this design. The point is to decouple the service generating proofs from the consumers of these proofs (and round data). In this design, the `Service` simply _produces_ round data and proofs. It has no knowledge about what happens to them afterward.

IMO it greatly simplifies testing as a consumer is now completely decoupled from the producer. They are connected with a channel only. Take a look at the unit tests. They don't need a mock for the hypothetical "database interface". It avoids creating untestable monsters like the atxBuilder in go-spacemesh where UTs are a mock hell.

service/db.go
fasmat: After discussing with @poszu we can leave the `ServiceDB` as it is and might take a look at it again in the future.

service/service.go
fasmat: see the other comment above.

fasmat (approved):

PR #245, opened 2023-02-22
Try to enable merge queues

pigmej:
Here are the docs: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions

That would be the proposed merge queue config.
Screenshot 2023-02-22 at 09 45 35

TBD is the timeout for starting the queue (5min vs 1min let's say).

Let's kinda use that repo as a working trial for these queues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ env:
   go-version: "1.19.6"
 
 on:
+  merge_group:
   pull_request:
   push:
     branches:


fasmat (approved):

fasmat (changes requested):
Approving this still allows to squash and merge without the queue?

EDIT: I guess we have to set the branch protections as well?

fasmat (approved):

pigmej:
@fasmat yes, I did not set it yet because first CI had to be adjusted. but yes we need to set what I have listed above in the screen.

PR #251, opened 2023-03-01
Expose method to obtain REST-GRPC proxy address

poszu:
It can be used by consumers of the poet (as a library) in tests to obtain am automitally-allocated REST proxy address.


fasmat (approved):

fasmat:
lgtm

PR #259, opened 2023-03-15
Specify max size of variable size types encoded with scale

fasmat:
This adds max sizes to all types serialized with scale.


shared/shared.go
```diff
@@ -75,26 +75,38 @@ func MakeLabelFunc() func(hash LabelHash, labelID uint64, leftSiblings [][]byte)
...
+	Value []byte `scale:"max=32"`
 }
 
 type MerkleProof struct {
-	Root         []byte
-	ProvenLeaves [][]byte
-	ProofNodes   [][]byte
+	Root         []byte `scale:"max=32"`
+	ProvenLeaves []Leaf `scale:"max=150"`
+	ProofNodes   []Node `scale:"max=5400"` // 36 nodes per leaf
```
poszu: Why can't it stay `[][]byte`? These new types generated a lot of additional boilerplate code.

fasmat: Because there is no simple way to specify the size of recursive slices:

```
type MyStruct struct {
    Field [][]byte `scale:"max=10"` // only specifies that len(Field) <= 10, not that len(Field[0]) <= 10
}

Comment (poszu) [reply shared/shared.go]: I see :(.

I gave it some more thought. I don't really see a point to export ("share") these structures by the Poet repo in the first place. The interface of the poet is GRPC and protobufs. If the user of the poet (i.e go-spacemesh) wants to convert proto structures into different formats, it should define structures to do so itself.

IMO, these structures shouldn't be "shared" and they shouldn't implement the scale interfaces at all. The only place where something is serialized is when `Proof` is saved in DB. For that, it needs neither consistency guarantees of scale nor size limitations.

File: shared/shared.go [hunk 2]
```diff
@@ -75,26 +75,38 @@ func MakeLabelFunc() func(hash LabelHash, labelID uint64, leftSiblings [][]byte)
...
+}
+
 type Proof struct {
 	// The actual proof.
 	MerkleProof
 
 	// Members is the ordered list of miners challenges which are included
 	// in the proof (by using the list hash digest as the proof generation input (the statement)).
-	Members [][]byte
+	Members []Member `scale:"max=150"`
```
poszu: Why 150? There could be much more members. A member is each node that submitted a challenge for the given round. In fact, there is no limit to how many members there can be in a round.

shared/shared.go
```diff
@@ -75,26 +75,38 @@ func MakeLabelFunc() func(hash LabelHash, labelID uint64, leftSiblings [][]byte)
...
+type Node struct {
+	Value []byte `scale:"max=32"`
 }
 
 type MerkleProof struct {
-	Root         []byte
-	ProvenLeaves [][]byte
-	ProofNodes   [][]byte
+	Root         []byte `scale:"max=32"`
+	ProvenLeaves []Leaf `scale:"max=150"`
```
poszu: This number should be in-sync with the difficulty parameter: https://github.com/spacemeshos/poet/blob/53f681ea6d46d571d38afa8be9a4f77997a87afe/shared/shared.go#L16

IMO, it should be at least documented.

shared/shared.go
fasmat: That would be an issue, since we have to specify some maximum number of members. What's a realistic upper limit? 1 Mio?

shared/shared.go
fasmat: This goes beyond the scope of updating to the most recent version of scale. The types in the `shared` package are used in `go-spacemesh` and they are encoded there with `go-scale` as well.

I agree that this shouldn't be the case but would move the refactoring of that into a separate issue.

poszu: But it complicates the existing code _a lot_. I mean code just converting from `[]byte` to the new structs, like:
```go
pLeaves := make([]shared.Leaf, 0, len(provenLeaves))
for _, leaf := range provenLeaves {
	pLeaves = append(pLeaves, shared.Leaf{
		Value: leaf,
	})
}
pNodes := make([]shared.Node, 0, len(proofNodes))
for _, node := range proofNodes {
	pNodes = append(pNodes, shared.Node{
		Value: node,
	})
}
```

How about doing it this way:
1. make a copy of the existing structures (the original ones) - "not-shared", use them inside of this repo - avoiding all the new boilerplate
2. keep the new structures (with added scale tags) in `package shared` for go-spacemesh to use
3. In follow up PRs move the structures from `shared` directly to go-spacemesh repo

It will limit the added boilerplate, allow you to quickly proceed with this PR, and clean the code up soon.

shared/shared.go
poszu: I'm not sure. 1M should be more than enough under normal conditions (unless the poet is spammed with challenges).

shared/shared.go
fasmat: This won't work. Tests in the `activation` package of `go-spacemesh` are calling `prover.GenerateProof` and `verifier.Validate` both of which return and expect a `shared.MerkleProof` which is also `serialized` with go-scale.

Either I do the conversation here or I have to do it every time I call one of these to functions in any place in `go-spacemesh`.

I don't think this code is convoluted or complicated, its a simple conversion from one type to another. Its just more verbose because there is no `Array.map(s -> t)` construct in go. As a compromise I implemented such a functional mapping but I don't see why I should now introduce 2 `shared.MerkleProof` types and convert between them 🤷

poszu: Which tests call `prover.GenerateProof`? I can't find them.

I looked through the code and as far as I can see, go-spacemesh (de)serializes `shared.MerkleProof` only. `shared.Proof` is (de)serialized in poet for storage in DB, scale was used out of convenience only - any serialization could be used, for example, `xdr` which is also already used in poet.

Bonus: Using something like `xdr` for Proof to save in DB would also remove the problem of deciding about the maximum number of members in Proof.

>I don't think this code is convoluted or complicated, its a simple conversion from one type to another.

It obscures the real logic in code w/o a good reason, or just because some application that depends on the poet's verification method also happens to serialize the Merkle proof for some reasons (reasons that the poet doesn't care about).

IMO the problem is that a requirement of go-spacemesh (that is - to be able to (de)serialize `shared.MerkleProof` leaks into the poet library).

fasmat: > Which tests call prover.GenerateProof? I can't find them.

All tests in `poetdb_tests.go` call this method indirectly via `prover.GenerateProofWithoutPersistency`

> Bonus: Using something like xdr for Proof to save in DB would also remove the problem of deciding about the maximum number of members in Proof.

Not really. If the members slice can be arbitrarily large we need to store an arbitrarily large object in our DB. We should limit the size independent of the encoder we use.

I would even go so far and say that if the maximum amount of users have submitted their challenges to poet, it should just refuse any additional request. We will have scaling issues if every user submits their challenge to every poet and every poet is supposed to store and process all of that data.

> IMO the problem is that a requirement of go-spacemesh (that is - to be able to (de)serialize `shared.MerkleProof` leaks into the poet library).

That is valid and needs to be cleaned up at some point, but is out of scope for this issue.

poszu: > Not really. If the members slice can be arbitrarily large we need to store an arbitrarily large object in our DB. We should limit the size independent of the encoder we use.
> 
> I would even go so far and say that if the maximum amount of users have submitted their challenges to poet, it should just refuse any additional request. We will have scaling issues if every user submits their challenge to every poet and every poet is supposed to store and process all of that data.

That is exactly how this should be enforced - on the API level, not during serialization. Otherwise, the poet will run for the entire epoch and fail miserably trying to save the proof in DB. It would be really bad.

I quickly refactored the existing code to not rely on scale for other structs than `shared.MerkleProof`: #272. Merging it first would limit footprint of this PR.

poszu (approved):

PR #268, opened 2023-03-20
Automatic releases

poszu:
Changes:
- CI now builds and tests with Go 1.20.2. Note that the minimum supported version in go.mod is still go1.19,
- Added workflow to automatically build poet binary and create a release when tag vx.y.z is pushed.


.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 name: CI
 
 env:
-  go-version: "1.19.7"
+  go-version: "1.20.2"
```
fasmat: I would prefer not specifying minor versions.

The only reason it exists is because of govulncheck; Go 1.20.3 comes out and it will mark all our builds as failing for about a week until GH updates the default version from Go 1.20.2 to Go 1.20.3

and this way we have to do this manually every 3-4 weeks.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.19-alpine as build
+FROM golang:1.20.2-alpine as build
```
fasmat: Same here, if 1.20.3 comes out 1.20 will point to that, 1.20.2 won't.

Minor releases are always only bugfixes, usually security relevant and shouldn't break any existing software.

fasmat (approved):

PR #272, opened 2023-03-22
Remove Proof and ProofMesage from package `shared`

poszu:
And don't implement scale encoding for them.

Bonus: added test for `service.ProofsDatabase`.


service/db_test.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func TestInsertAndGetProof(t *testing.T) {
+	tempdir := t.TempDir()
+	proofs := make(chan service.ProofMessage)
+	db, err := service.NewProofsDatabase(tempdir, proofs)
+	require.NoError(t, err)
+
+	var eg errgroup.Group
+	ctx, cancel := context.WithCancel(context.Background())
```
fasmat, line 19: Just in case a test exits early because of a failed require:

```golang
defer cancel()
```

so any running go-routines get the signal to stop 🙂

service/types.go
```diff
@@ -0,0 +1,21 @@
+package service
+
+import "github.com/spacemeshos/poet/shared"
+
+type Proof struct {
```
fasmat: If these types are not intended to be used outside of the `service` package, lets make them private.

fasmat (approved):

PR #277, opened 2023-03-30
PoW for /Submit and dropped integration with Gateway

### LINKED ISSUE #274: Poet challenge PoW

poszu:
Closes #274


shared/shared.go
```diff
@@ -171,3 +173,55 @@ func DecodeSliceOfByteSliceWithLimit(d *scale.Decoder, outerLimit, innerLimit ui
...
+
+		hash := CalcSubmitPowHash(powChallenge, poetChallenge, nodeID, hash, nonce)
+		if CheckLeadingZeroBits(hash, difficulty) {
+			return nonce, nil
+		}
+	}
+}
+
+// CalcSubmitPowHash calculates the hash for the Submit PoW.
+// The hash is ripemd256(sha256(powChallenge || poetChallenge || nodeID || nonce).
```
dshulyak: > ripemd256(sha256(powChallenge || poetChallenge || nodeID || nonce)

it doesn't seem accurate, is it outdated? it looks like sha256 should be removed (also in other places)

dshulyak (approved):

shared/shared.go
poszu: thx

PR #284, opened 2023-04-12
Disable vulncheck

fasmat:
Unless we want to update our CI configuration every 3-4 weeks with a new minor version of go to prevent vulncheck from marking them as vulnerable we should temporarily disable the tool. From the official documentation:

"Govulncheck is an experimental tool under active development."

As long as it is experimental it might cause more issues than actual helpful feedback for us.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 
 env:
-  go-version: "1.20.4"
+  go-version: "1.20"
 
 on:
   merge_group:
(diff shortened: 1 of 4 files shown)


piersy:
@fasmat Why is this PR also downgrading go and upgrading clock?

fasmat:
> @fasmat Why is this PR also downgrading go and upgrading clock?

It's not downgrading go. It just removes the specification for the minor version so we don't have to update it every 3-4 weeks 😊.

The clock update is unrelated and I can remove it from the PR if you want.

fasmat:
You can see here that for the last CI run it used 1.20.4: https://github.com/spacemeshos/poet/actions/runs/5092432992/jobs/9154300559

fasmat:
superseded by https://github.com/spacemeshos/poet/pull/284

PR #292, opened 2023-05-15
Use blake3 for membership merkle tree and prefix to mitigate second p…

poszu:
…reimage attack

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 
 env:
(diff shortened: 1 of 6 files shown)


dshulyak (approved):
this is only for membership, right? e.g sha256 is used for poet proof computation

poszu:
> this is only for membership, right? e.g sha256 is used for poet proof computation

Yes. The poet PoSW tree uses different hashing functions for leafs and nodes.

PR #297, opened 2023-05-23
Offload round teardown to a background goroutine

### LINKED ISSUE #296: Poet-baremetal requires serious time to write proof to a disk
image

Merkle tree generation finished at 170000, proof ended 170007 and yet somehow proof was written to a database at 170049.

The other BM poet needed 58 seconds.

poszu:
Closes #296 

Apparently, round teardown takes a very long time (40-50s) on the testnet poets. The teardown removes lots of data from the disk. This PR mitigates it by offloading round teardown onto a goroutine running in the background. The goroutine is placed in errgroup that is awaited at the end of the `Service` loop to make sure it's completed before shutdown.

todo:
- [x] cleanup logs
- [x] offload round teardown onto a background goroutine

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -229,7 +229,6 @@ func (s *Service) loop(ctx context.Context, roundToResume *round) error {
 			unlock := lockOSThread(ctx, roundTidFile)
 			defer unlock()
 			err := round.recoverExecution(ctx, end, s.cfg.TreeFileBufferSize)
-			logger.Info("recovered round execution finished", zap.String("round", round.ID))
 			roundResults <- roundResult{round: round, err: err}
 			return nil
 		})
@@ -242,15 +241,16 @@ func (s *Service) loop(ctx context.Context, roundToResume *round) error {
 
 		case result := <-roundResults:
 			if result.err == nil {
-				s.onNewProof(ctx, result.round.ID, result.round.execution)
+				s.onNewProof(result.round.ID, result.round.execution)
 			} else {
 				logger.Error("round execution failed", zap.Error(result.err), zap.String("round", result.round.ID))
 			}
-			if err := result.round.teardown(result.err == nil); err != nil {
-				logger.Warn("round teardown failed", zap.Error(err))
-			} else {
-				logger.Info("round teardown finished", zap.String("round", result.round.ID))
-			}
+			eg.Go(func() error {
+				if err := result.round.teardown(result.err == nil); err != nil {
+					logger.Warn("round teardown failed", zap.Error(err), zap.String("round", result.round.ID))
+				}
+				return nil
+			})
 			delete(s.executingRounds, result.round.ID)
 
 		case <-s.timer:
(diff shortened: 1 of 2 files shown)


service/service.go
```diff
@@ -280,7 +282,25 @@ func (s *Service) loop(ctx context.Context, roundToResume *round) error {
 			if err := s.openRound.teardown(false); err != nil {
 				return fmt.Errorf("tearing down open round: %w", err)
 			}
-			return nil
+			_ = eg.Wait()
```
fasmat, line 280:
```suggestion
			eg.Wait()
```

poszu: `_ = eg.Wait()` **explicitly** (on purpose) swallows an error informing that we are not interested in the error here. Plus, linter would yell :)

fasmat: Then the linter configuration imo should be changed, `golangci-lint` does not complain if `_ =` is removed (unless `gosec` has been activated).

`eg.Wait()` will never return an error, because no routine started with it returns anything but `nil` 😉. 

`gosec` is inconsistent in its reporting. This is ok:

```go
fmt.Println("Hello, World!")
```

although should be:

```go
_, _ = fmt.Println("Hello, World!")
```

Imo we don't need to **explicitly** ignore every error - its clear from the code that `eg.Wait()` doesn't need to be checked.

poszu: > eg.Wait() will never return an error, because no routine started with it returns anything but nil :wink:.

Yes, that's why it's fine to ignore the error with `_ = `. I can't do anything with `gosec` being inconsistent. Perhaps it doesn't detect an unhandled error if there is more than 1 return value?

pigmej:
@poszu In the same time I think we should find a proper solution also.

poszu:
> @poszu In the same time I think we should find a proper solution also.

I'm not sure if there is anything to do on the implementation side. The code simply calls [`os.RemoveAll()`](https://pkg.go.dev/os#RemoveAll). If it takes way longer than expected, the reason is probably on OS/HW side.

service/service.go
fasmat: I think maybe using `errgroup` here is why the workaround is needed in the first place. Maybe `sync.Waitgroup` would be a better fit?

`errgroup` is basically a wrapper around `sync.Waitgroup` to extend it with error propagation, and Context cancelation. If that is not needed, maybe the Waitgroup would work instead?

```golang
var wg sync.Waitgroup
wg.Add(1)
go func() {
	defer wg.Done()
	// some task
	// no return nil
}

wg.Wait() // doesn't return an err
```

poszu: I avoid Waitgroup because it requires manually calling `.Add()` and `.Done()` which is error-prone.

fasmat: I tend to disagree. There isn't much difference between

```golang
eg.Go(func() error {
	// task
	return nil
})
eg.Wait()
```

and

```golang
wg.Add(1)
go func() {
	defer wg.Done()
	// task
}()
wg.Wait()
```

it also has the advantage of no need to explicitly ignore the error handling that we aren't using anyway (as well as built in cancelation). Its also easier to scope variables if needed.

poszu: IMHO, there is a huge difference. In the second snippet, it's easy to forget `wg.Add()` or `wg.Done()` and create buggy code. This is not possible with the errgroup.

> Its also easier to scope variables if needed.

I don't know what you mean.

fasmat (approved):

pigmej:
Ok will think about something on OS level. Maybe there is some tuning that we could do for that scenario.

PR #299, opened 2023-05-24
Remove appdata.go and tweak config file reading

poszu:


piersy:
Hi @poszu there seems to be a fair amount of code dedicated here to supporting multiple default paths for config. What about just making the user explicitly provide config with a flag?

poszu:
> Hi @poszu there seems to be a fair amount of code dedicated here to supporting multiple default paths for config. What about just making the user explicitly provide config with a flag?

True it will look at a few places. However, there is a flag to provide path to a config file. If not set it will look in the:
- current directory,
- in OS config directory (for example, `~/.config/poet/poet.conf` on linux).

Is it necessary? Probably not. @nj, you are the main user of the poet - what is important to you? I can make it look for config **only** when `-c <path>` is given on CLI and remove looking in the aforementioned dirs.

pigmej:
I don't think it matters that much for me honestly :)) I provide the config path anyway because it's in different place anyway ;)

fasmat (approved):

fasmat:
I would also prefer the user to have to provide paths explicitly instead of using too much logic to define a default path.

Normally I would say if `os.UserCacheDir()` errors, the app should just log the error and `os.Exit(1)`, but since this is executed even when a custom path is set that would be counterproductive and changing the code to only determine the default path when no custom path is set is probably a bigger effort?

PR #300, opened 2023-05-24
Refactor executingRounds as there's only 1 ever executed at a time

poszu:
GRPC is left unmodified (it still returns an array) as it would break clients already using this API.


rpc/rpcserver.go
```diff
@@ -96,26 +95,21 @@ func (r *rpcServer) Info(ctx context.Context, in *api.InfoRequest) (*api.InfoRes
...
+	}
 
-	ids := make([]string, len(info.ExecutingRoundsIds))
-	copy(ids, info.ExecutingRoundsIds)
-	out.ExecutingRoundsIds = ids
-	out.ServicePubkey = r.s.PubKey
-	out.PhaseShift = durationpb.New(r.cfg.Service.PhaseShift)
-	out.CycleGap = durationpb.New(r.cfg.Service.CycleGap)
+	if info.ExecutingRoundId != nil {
+		out.ExecutingRoundsIds = []string{*info.ExecutingRoundId}
```
fasmat: I know changing it to only one round would break the API of poet, but is it already to late to change it?

fasmat (approved):

rpc/rpcserver.go
poszu: Probably not too late, but such change can't go into the testnet. I could add a new field in proto `ExecutingRoundId string` and deprecate the plural version. Then after go-spacemesh migrates I could remove the old field (before genesis). Wdyt?

fasmat: sounds reasonable to me 👍

poszu: As the testnet is gone anyway and a new one will be set up, I could fit the change in GRPC

PR #302, opened 2023-05-30
Avoid manual updates for go version

fasmat:
Instead of specifying the exact version of go to use when building `poet` use the version specified in `go.mod`.

Additionally always check if the version used is the newest patch release of the version specified in `go.mod` to avoid `vulncheck` to complain about an outdated release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: CI
 
+env:
+  go-version: "1.20"
+
 on:
   merge_group:
   pull_request:
@@ -18,7 +21,7 @@ jobs:
         uses: actions/setup-go@v4
         with:
           check-latest: true
-          go-version-file: "go.mod"
+          go-version: ${{ env.go-version }}
       - name: fmt, tidy, generate
         run: |
           make install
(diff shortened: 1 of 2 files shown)


poszu:
I'm not sure. AFAIR, the version specified in go.mod is the minimum supported version (the minimum Go version required to build). It's not the version that we want to build a release binary. We want to build a release binary with go 1.20 because it results in a slightly more performant binary.

fasmat:
We can still use `check-latest` to ensure that the newest patch version is used (and vulncheck doesn't complain) and specify to use 1.20 with a env-variable instead.

piersy:
This seems reasonable to me.

@poszu I'm not really sure what your point is, AFAICS this PR is just ensuring that actions uses the latest 1.20.xxx go version, which will keep vulncheck happy.

fasmat:
@piersy before I didn't set 1.20 explicitly but instead used

```yaml
go-version-file: "go.mod"
```

as version constraint, which would use 1.19 to build and test the code, since that's the version specified in `go.mod`. We still have 1.19 there because `go-spacemesh` still uses 1.19 and depends on `poet`, but `poet` releases are built using 1.20 because of better performance.

piersy:
@fasmat Thanks for the clarification. This looks good to me, but I'll let @poszu have the final word.

poszu (approved):

pigmej (approved):

PR #312, opened 2023-06-26
Refactor Genesis type in Config to time.Time

poszu:
Changed `Genesis` type from `string` to `type Genesis time.Time`, which implements `flags.Unmarshaler` interface from go-flags for automatic parsing. Therefore, manual parsing in code is not required anymore.


fasmat (approved):
`time.Time` should really be a supported type by the package we are using for parsing; `time.Duration` is already.

Unfortunately it seems like `go-flags` hasn't been updated in 5 years, so it is unlikely that this will be added in the near future.

poszu:
> `time.Time` should really be a supported type by the package we are using for parsing; `time.Duration` is already.
> 
> Unfortunately it seems like `go-flags` hasn't been updated in 5 years, so it is unlikely that this will be added in the near future.

Then the author would have to pick specific time format (or make it somehow configurable).

PR #326, opened 2023-07-26
avoid panic when Validate

zhiqiangxu:
Currently an evil node can generate proofs that make the node panic when [`verifier.Validate`](https://github.com/spacemeshos/poet/blob/0ba4bc7e8ea47d9f5371da27ef5a898173aeb6df/verifier/verifier.go#L16) is called.

[Here](https://github.com/zhiqiangxu/spacemesh-tool) is a proof of concept to demonstrate this:

just clone the repo and run `go run main.go poc poet`, then you'll see it panics inside `verifier.Validate`.

 
image

This PR tries to fix this issue to avoid potential dos problem.

### DIFF
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -20,6 +20,17 @@ func Validate(proof shared.MerkleProof, labelHashFunc func(data []byte) []byte,
 		return fmt.Errorf("number of proven leaves (%d) must be equal to security param (%d)",
 			len(proof.ProvenLeaves), securityParam)
 	}
+	if len(proof.ProvenLeaves)*36 < len(proof.ProofNodes) {
+		return fmt.Errorf("for every proven leaf (%d) there must be at most 36 proof nodes (%d)",
+			len(proof.ProvenLeaves), len(proof.ProofNodes),
+		)
+	}
+	if len(proof.ProvenLeaves) > len(proof.ProofNodes) {
+		return fmt.Errorf("for every proven leaf (%d) there must be at least 1 proof node (%d)",
+			len(proof.ProvenLeaves), len(proof.ProofNodes),
+		)
+	}
+
 	provenLeafIndices := asSortedSlice(shared.FiatShamir(proof.Root, numLeaves, securityParam))
 	provenLeaves := make([][]byte, 0, len(proof.ProvenLeaves))
 	for i := range proof.ProvenLeaves {
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -43,6 +43,9 @@ func Validate(proof shared.MerkleProof, labelHashFunc func(data []byte) []byte,
 		return fmt.Errorf("merkle proof not valid")
 	}
 
+	if len(parkingSnapshots) != len(proof.ProvenLeaves) {
+		return fmt.Errorf("merkle proof incomplete")
+	}
 	makeLabel := shared.MakeLabelFunc()
 	for id, label := range proof.ProvenLeaves {
 		expectedLabel := makeLabel(labelHashFunc, provenLeafIndices[id], parkingSnapshots[id])
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -55,7 +55,7 @@ func Validate(proof shared.MerkleProof, labelHashFunc func(data []byte) []byte,
 	}
 
 	if len(parkingSnapshots) != len(proof.ProvenLeaves) {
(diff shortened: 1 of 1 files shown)


fasmat:
Hi @zhiqiangxu!

Thanks for your contribution! The `verifier.Validate` function is indeed missing checks to ensure that it doesn't panic. A valid proof will have at least 1 and at most 36 nodes per leaf. Could you add the following checks to the `Validate` function (after the check for `securityParam` == `len(proof.ProvenLeaves)`)?:

```go
	if len(proof.ProvenLeaves)*36 < len(proof.ProofNodes) {
		return fmt.Errorf("for every proven leaf (%d) there must be at most 36 proof nodes (%d)",
			len(proof.ProvenLeaves), len(proof.ProofNodes),
		)
	}
	if len(proof.ProvenLeaves) > len(proof.ProofNodes) {
		return fmt.Errorf("for every proven leaf (%d) there must be at least 1 proof node (%d)",
			len(proof.ProvenLeaves), len(proof.ProofNodes),
		)
	}
```

This should prevent the panic. The reason your example program causes the crash is because the proof doesn't contain any `ProofNodes` thereby causing `merkle.ValidatePartialTreeWithParkingSnapshots` to return an invalid result.

fasmat (changes requested):
Hi @zhiqiangxu!

Thanks for the investigation and the suggested fix 🙂 !

The `verifier.Validate` function is indeed missing checks to ensure that it doesn't panic. A valid proof will have at least 1 and at most 36 nodes per leaf. Could you add the following checks to the `Validate` function (after the check for `securityParam` == `len(proof.ProvenLeaves)`)?:

```go
	if len(proof.ProvenLeaves)*36 < len(proof.ProofNodes) {
		return fmt.Errorf("for every proven leaf (%d) there must be at most 36 proof nodes (%d)",
			len(proof.ProvenLeaves), len(proof.ProofNodes),
		)
	}
	if len(proof.ProvenLeaves) > len(proof.ProofNodes) {
		return fmt.Errorf("for every proven leaf (%d) there must be at least 1 proof node (%d)",
			len(proof.ProvenLeaves), len(proof.ProofNodes),
		)
	}
```

This should prevent the panic. The reason your example program causes the crash is because the proof doesn't contain any `ProofNodes` thereby causing `merkle.ValidatePartialTreeWithParkingSnapshots` to return an invalid result.

zhiqiangxu:
> Hi @zhiqiangxu!
> 
> Thanks for the investigation and the suggested fix 🙂 !
> 
> The `verifier.Validate` function is indeed missing checks to ensure that it doesn't panic. A valid proof will have at least 1 and at most 36 nodes per leaf. Could you add the following checks to the `Validate` function (after the check for `securityParam` == `len(proof.ProvenLeaves)`)?:
> 
> ```go
> 	if len(proof.ProvenLeaves)*36 < len(proof.ProofNodes) {
> 		return fmt.Errorf("for every proven leaf (%d) there must be at most 36 proof nodes (%d)",
> 			len(proof.ProvenLeaves), len(proof.ProofNodes),
> 		)
> 	}
> 	if len(proof.ProvenLeaves) > len(proof.ProofNodes) {
> 		return fmt.Errorf("for every proven leaf (%d) there must be at least 1 proof node (%d)",
> 			len(proof.ProvenLeaves), len(proof.ProofNodes),
> 		)
> 	}
> ```
> 
> This should prevent the panic. The reason your example program causes the crash is because the proof doesn't contain any `ProofNodes` thereby causing `merkle.ValidatePartialTreeWithParkingSnapshots` to return an invalid result.

Done, but curious where is the max height 36 guaranteed?

fasmat:
Thank you!

Theoretically there can be more than 36 nodes per leaf, but at the moment a proof of that size wouldn't be able to be encoded/decoded using scale: https://github.com/spacemeshos/poet/blob/0ba4bc7e8ea47d9f5371da27ef5a898173aeb6df/shared/shared.go#L81-L87

verifier/verifier.go
```diff
@@ -43,6 +54,9 @@ func Validate(proof shared.MerkleProof, labelHashFunc func(data []byte) []byte,
 		return fmt.Errorf("merkle proof not valid")
 	}
 
+	if len(parkingSnapshots) != len(proof.ProvenLeaves) {
+		return fmt.Errorf("merkle proof incomplete")
```
fasmat:
```suggestion
		return fmt.Errorf("merkle proof not valid")
```

PR #327, opened 2023-07-26
Remove incorrect checks for valid proof

fasmat:
It seems like the assumption that there needs to be at least one node per leaf wasn't correct, system tests will fail with these checks so I removed them again.

Additionally parallelisation of tests has been removed. `make test` already executed only one test in parallel with `-p 1` if removed tests will become even more flaky. Instead I removed `t.Parallel()` from the tests to prevent them being executed in parallel accidentally.

### DIFF
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -18,16 +18,7 @@ func Validate(proof shared.MerkleProof, labelHashFunc func(data []byte) []byte,
 ) error {
 	if int(securityParam) != len(proof.ProvenLeaves) {
 		return fmt.Errorf("number of proven leaves (%d) must be equal to security param (%d)",
-			len(proof.ProvenLeaves), securityParam)
-	}
-	if len(proof.ProvenLeaves)*36 < len(proof.ProofNodes) {
-		return fmt.Errorf("for every proven leaf (%d) there must be at most 36 proof nodes (%d)",
-			len(proof.ProvenLeaves), len(proof.ProofNodes),
-		)
-	}
-	if len(proof.ProvenLeaves) > len(proof.ProofNodes) {
-		return fmt.Errorf("for every proven leaf (%d) there must be at least 1 proof node (%d)",
-			len(proof.ProvenLeaves), len(proof.ProofNodes),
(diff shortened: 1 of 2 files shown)


poszu:
> Additionally parallelisation of tests has been removed. `make test` already executed only one test in parallel with `-p 1` if removed tests will become even more flaky. Instead I removed `t.Parallel()` from the tests to prevent them being executed in parallel accidentally.

With `-p 1`, go will not run tests from many _packages_ at the same time, but tests marked with `t.Parallel()` that are located in the same package can still be executed in parallel.

verifier/verifier.go
```diff
@@ -18,16 +18,7 @@ func Validate(proof shared.MerkleProof, labelHashFunc func(data []byte) []byte,
 ) error {
 	if int(securityParam) != len(proof.ProvenLeaves) {
 		return fmt.Errorf("number of proven leaves (%d) must be equal to security param (%d)",
-			len(proof.ProvenLeaves), securityParam)
-	}
-	if len(proof.ProvenLeaves)*36 < len(proof.ProofNodes) {
-		return fmt.Errorf("for every proven leaf (%d) there must be at most 36 proof nodes (%d)",
-			len(proof.ProvenLeaves), len(proof.ProofNodes),
-		)
-	}
```
poszu, line 27: This check should be fine, shouldn't it? The same is enforced in scale limits.

poszu (approved):

verifier/verifier.go
fasmat: Scale limits enforce that no more than 5400 nodes are contained in a proof and 150 leaves. I'm not sure if > 36 nodes can exist per leave, but there definitely can be less than 1. 

With both checks in place all unit tests pass, except for mac which sometimes fails because there are fewer nodes than leaves. System tests also fail with the checks in place.

I didn't want to assume anything about how the trees are constructed if this change already lead to random failures and instead only keep the check that avoids a panic. 🙂

poszu: There can be fewer nodes per leaf because for some leaves the paths from a leaf to the root of the tree can overlap.

PR #338, opened 2023-08-07
Add more logs and round members metric

### LINKED ISSUE #336: Improve observability around challenge registration
- log the number of registered challenges when a round starts,
- log the state of a recovered round (when poet is rebooted),
- provide a metric to view the number of registered challenges,

poszu:
Closes #336 

- added some logs, especially log the number of round members when it's recovered and started 
- refactored existing logs
- improved logging around context cancellation (i.e. previously it incorrectly logged _round failure_ when service was stopped)
- added metric showing the number of members
- refactored leaves counter metric
  ```
  # HELP poet_round_leaves_total Number of generated leaves in a round
  # TYPE poet_round_leaves_total counter
  poet_round_leaves_total{ID="2"} 1.478935e+06
  poet_round_leaves_total{ID="3"} 0
  # HELP poet_round_members_total Number of members in a round
  # TYPE poet_round_members_total counter
  poet_round_members_total{ID="2"} 0
  poet_round_members_total{ID="3"} 1
  ```

### DIFF
--- a/service/round.go
+++ b/service/round.go
@@ -32,14 +32,14 @@ var (
 		Subsystem: "round",
 		Name:      "members_total",
 		Help:      "Number of members in a round",
-	}, []string{"ID"})
+	}, []string{"id"})
 
 	leavesMetric = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "poet",
 		Subsystem: "round",
 		Name:      "leaves_total",
 		Help:      "Number of generated leaves in a round",
-	}, []string{"ID"})
+	}, []string{"id"})
 )
 
 type executionState struct {
--- a/service/round.go
+++ b/service/round.go
@@ -7,10 +7,10 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
-	"sync/atomic"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/spacemeshos/merkle-tree"
 	"github.com/spacemeshos/merkle-tree/cache"
 	"github.com/syndtr/goleveldb/leveldb"
@@ -26,6 +26,13 @@ import (
 var (
 	ErrRoundIsNotOpen    = errors.New("round is not open")
 	ErrMaxMembersReached = errors.New("maximum number of round members reached")
+
+	membersMetric = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "poet",
+		Subsystem: "round",
+		Name:      "members_total",
+		Help:      "Number of members in a round",
+	}, []string{"ID"})
 )
 
 type executionState struct {
@@ -42,7 +49,6 @@ const roundStateFileBaseName = "state.bin"
 type roundState struct {
 	ExecutionStarted time.Time
 	Execution        *executionState
-	Members          uint
 }
 
 func (r *round) isOpen() bool {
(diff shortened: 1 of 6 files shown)


service/round.go
```diff
@@ -271,13 +303,62 @@ func (r *round) calcMembersAndStatement() ([][]byte, []byte, error) {
...
+}
+
+func newRoundsMetricCollector(round *round) prometheus.Collector {
+	return &roundMetricCollector{
+		round: round,
+		totalMembersDesc: prometheus.NewDesc(
+			prometheus.BuildFQName("poet", "round", "members_total"),
+			"the total number of members in a round",
+			nil,
+			prometheus.Labels{"epoch": round.ID},
```
dshulyak: this way it doesn't report unregistered rounds?

fasmat: Why the wrapper around the collector to access round upon a request?

I think it would be simpler to create a counter similar to how the leavesCounter is set up at the moment and call Inc() on it whenever a new member successfully submitted.

This would also remove the need to make members an atomic.uint64

fasmat: To have the metric with a label you probably need CounterVec instead of Counter.

Initialize and register it when the server starts and deregister it upon shutdown.  It can then be updated by calling `memberMetric.WithLabelValues(roundID).Add(1)` in the `submit` method.

dshulyak: this approach doesn't work if set of labels is unbounded, as prometheus client will keep them in memory until restarted

poszu: @dshulyak 
>this way it doesn't report unregistered rounds?

Yes. The key was to have distinct `Desc`.

@fasmat 
> To have the metric with a label you probably need CounterVec instead of Counter.

As @dshulyak said, prometheus doesn't unregister the `Counter`s and it keeps growing. It grows 1 counter/2 weeks but still.

The approach with a `Collector` seems overcomplicated to me. I'm considering changing it to just a `Counter` with a const label for the epoch. Unregistering works fine (if `CounterVec` is **not** used).

poszu: @dshulyak, @fasmat I managed to get the `CounterVec` to work without leaking past Counters. There is a `CounterVec::Delete` method to remove a Counter from it.

Refactored the code to use `CounterVec` - it's much simpler now.

service/round.go
```diff
@@ -24,6 +26,20 @@ import (
 var (
 	ErrRoundIsNotOpen    = errors.New("round is not open")
 	ErrMaxMembersReached = errors.New("maximum number of round members reached")
+
+	membersMetric = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "poet",
+		Subsystem: "round",
+		Name:      "members_total",
+		Help:      "Number of members in a round",
+	}, []string{"ID"})
```
dshulyak: could you please use lowercased values?

poszu: sure

service/round.go
```diff
@@ -76,22 +94,36 @@ func newRound(datadir string, epoch uint32, maxMembers uint) (*round, error) {
 		return nil, err
 	}
 
-	return &round{
+	// Note: using the panicking version of prometheus.NewCounterVec() because it panics
```
poszu:
```suggestion
	// Note: using the panicking version here because it panics
```

service/round.go
```diff
@@ -104,13 +136,15 @@ func (r *round) submit(key, challenge []byte) error {
 	err := r.challengesDb.Put(key, challenge, &opt.WriteOptions{Sync: true})
 	if err == nil {
 		r.members += 1
+		if r.membersCounter != nil {
+			r.membersCounter.Inc()
+		}
```
poszu:
```suggestion
		r.membersCounter.Inc()
```

fasmat (approved):

PR #341, opened 2023-08-08
Reject conficting challenge registrations

### LINKED ISSUE #340: Reject second registration with a different challenge
The /Submit endpoint is idempotent in a sense that it returns 200 OK if a given nodeID wants to submit a challenge for the second time in a round. However, it only checks if particular nodeID is already registered and does not compare the challenges (the new one vs the one in DB).

## Acceptance criteria
When the same nodeID submits a challenge again in the same round, the poet should return:
- 409 Conflict when the challenge is **different** from the one already registered,
- 200 OK when the challenge is identical to the one already registered.

poszu:
Closes #340

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -323,6 +323,45 @@ func TestCannotSubmitMoreThanMaxRoundMembers(t *testing.T) {
 	)
 }
 
+// Test if cannot submit two different challenges with the same nodeID.
+func TestSubmittingChallengeTwice(t *testing.T) {
+	t.Parallel()
+	req := require.New(t)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	cfg := config.DefaultConfig()
+	cfg.PoetDir = t.TempDir()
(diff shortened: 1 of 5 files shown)


server/server_test.go
```diff
@@ -325,6 +325,48 @@ func TestCannotSubmitMoreThanMaxRoundMembers(t *testing.T) {
...
+
+	cfg := config.DefaultConfig()
+	cfg.PoetDir = t.TempDir()
+	cfg.RawRPCListener = randomHost
+	cfg.RawRESTListener = randomHost
+
+	srv, client := spawnPoet(ctx, t, *cfg)
+
+	var eg errgroup.Group
+	eg.Go(func() error {
```
fasmat, line 343: The `errgroup.Group` isn't awaited. Maybe it makes sense to add this?

```go
t.Cleanup(func() { assert.NoError(t, eg.Wait()) } )
```

fasmat (approved):

PR #343, opened 2023-08-11
Add option to store DBs in a different location

### LINKED ISSUE #324: Provide way to separate leveldb and poet tree on the FS level
Currently they both reside on the same path. It would be nice to make them separate.

### LINKED ISSUE #331: Keep round's members indefinitely
# motivation
Currently, the list of members of the previous round is not persisted otherwise than with the proof (and available to fetch via /proofs endpoint). Members are kept in a per-round leveldb database (each round has its own DB). The DB is removed after the round is finished. 

As a prerequisite for #200, we should keep present and past rounds' members indefinitely. 

# design proposal
- centralize round's DB - use a single database shared by all rounds
- each round would have its unique DB key prefix, i.e `round-0-` to be able to query members for a given round.
- create a centralized service for round management (querying members and submitting challenges to the open round) - move the logic out of the poet worker service

poszu:
Closes #324 

Added `dbdir` config option (`--dbdir` on CLI) to configure the locations where databases (proofs and rounds DBs) are stored. The default value is `poetdir/db`.

The existing databases are automatically migrated out from `datadir`, where they resided before. 

The `dbdir` is never deleted, hence this pr also closes #331 allowing for indefinite storage of round members.

### `datadir`  structure:
:bulb: Note that `rounds/0` is missing.
```sh
├── proofs
│   ├── 000001.log
│   ...
├── round.tid
├── rounds
│   ├── 1
│   │   ├── challengesDb
│   │   │   ├── 000001.log
│   │   │    ...
│   │   ├── layercache_0.bin
│   │   └── state.bin
│   └── 2
│       ├── challengesDb
│       │   ├── 000001.log
│       │   ...
│       └── state.bin
└── state.bin
```

### `datadir`  structure:
```sh
├── round.tid
├── rounds
│   ├── 1
│   │   ├── layercache_0.bin
│   │   └── state.bin
│   └── 2
│       └── state.bin
└── state.bin
```

###  `dbdir` directory:
:bulb: Note that `rounds/0` is present.
```sh
├── proofs
│   ├── 000002.ldb
│   ...
└── rounds
    ├── 0
    │   ├── 000001.ldb
    │   ...
    ├── 1
    │   ├── 000001.ldb
    │   ...
    └── 2
        ├── 000001.ldb
       ...
```


server/server.go
```diff
@@ -122,7 +123,14 @@ func (s *Server) Start(ctx context.Context) error {
 		}),
 	}
 
-	proofsDbPath := filepath.Join(s.cfg.DataDir, "proofs")
+	proofsDbPath := filepath.Join(s.cfg.DbDir, "proofs")
+	// Temporary migration to move DB to a new location.
+	// TODO(poszu): remove this code after all poets are migrated.
+	oldProofsDbPath := filepath.Join(s.cfg.DataDir, "proofs")
```
fasmat: I think we should also mention this in the release notes, otherwise someone operating their own PoET might miss it.

We should also mention it again when this is removed and that new layout is now expected.

service/round.go
```diff
@@ -110,12 +126,12 @@ func newRound(datadir string, epoch uint32, maxMembers uint) (*round, error) {
 		execution: &executionState{
 			SecurityParam: shared.T,
 		},
-		members:        countMembersInDB(db),
 		maxMembers:     maxMembers,
 		membersCounter: membersCounter,
 		leavesCounter:  leavesCounter,
 	}
 
+	r.members = countMembersInDB(r.challengesDb)
```
fasmat: Why this change? Seems to be functional equivalent to the previous code.

fasmat (commented):
Would it make sense to have a CLI flag that only migrates the data from old to new and exits? Then the PoET code itself can just assume the new layout is already present and and the code for the migration is separate from the code of the service.

service/round.go
poszu: IDK, probably an artifact of development process. Reverted back to the original.

server/server.go
poszu: Sure, I will add information about the new configuration option to specify a different database directory to release notes.

~It defaults to the original behaviour (`DbDir` defaults to `DataDir`). If somebody is unware and runs without specifing the former, then everyting will work as before.~ The migration code is there only to help make the transition **only once** - it will not migrate if `DbDir` is changed from A to B.

Edit: changed the default `dbdir` path to `poetdir/db`. Everyone who upgrade will migrate their DBs.

poszu:
> Would it make sense to have a CLI flag that only migrates the data from old to new and exits? Then the PoET code itself can just assume the new layout is already present and and the code for the migration is separate from the code of the service.

I'm not sure, doing it automatically is easier to deploy, especially considering a cloud deployment. Now one can just add a `--dbdir=<path>` and mount a new PVC at that <path> and update the config and it will migrate automatically. Doing it as an explicit cmdline option would require some manual work.

@pigmej Wdyt?

pigmej:
Yes, please no migration INSIDE of POET, let's assume that it has only valid location. Please also note that for cloud deployments it should accept path in the same PVC as actual data.

poszu:
Actually, I'm now thinking of not defaulting `DbDir` to `DataDir` and making the transition opt-out (instead opt-in as right now). It should be safe to default `DbDir=`PoetDir/db`.

Migrating the round databases to a new location is important to keep the DB after a round is finished (the rounds/<ID> directory is removed) (#331).

poszu:
@pigmej 
> Yes, please no migration INSIDE of POET, let's assume that it has only valid location. Please also note that for cloud deployments it should accept path in the same PVC as actual data.

I don't understand your answer. Could you rephrase?

pigmej:
Let's assume that poet gets everything as needed. No `--migrate` flag.

poszu:
> Let's assume that poet gets everything as needed. No `--migrate` flag.

Then the maintainer will be responsible for a sequence:
- kill poet
- migrate DBs to new location
- restart the new version of poet and specify --dbdir

The maintainer will also need to understand the structure of directories under --dbdir:
```
dbdir
├── proofs
│   ├── 000002.ldb
│   ...
└── rounds
    ├── 0
    │   ├── 000001.ldb
    │   ...
```

Then `--dbdir` must **not** default to `--datadir` as the directory structure changes a bit (missing intermediate `challengesDb` layer under `rounds/x`)

I have a feeling it will cause problems... How will we update the cloud poets?

Edit: I could write a small migration utility to do this.

fasmat (approved):

pigmej (approved):

PR #344, opened 2023-08-15
Recovering lost parked nodes

poszu:
Recover parked nodes and the top layer of the tree from the layercache.bin files saved on disk instead from the round's state.bin.

## Parked nodes recovery
Parked nodes on disk-cached layers are recovered from the files (representing the tree's bottom layers). If a layer contains an odd number of nodes/leafs, the last one is a parked node.

Parked nodes on mem-cached layers are recovered by rebuilding a subtree, in which the bottom layer is built from the nodes in the top disk-cached layer.

## mem-cached nodes recovery
The top of the tree that is not persisted on disk is rebuilt as the aforementioned subtree. It's layers from 1 to the root are moved into the recovered tree.

TODO:
- [x] remove parked nodes completely from the round execution state and rely solely on recovery implemented in this PR.

### DIFF
--- a/prover/prover.go
+++ b/prover/prover.go
@@ -76,10 +76,9 @@ func GenerateProofRecovery(
 	limit time.Time,
 	securityParam uint8,
 	nextLeafID uint64,
-	parkedNodes [][]byte,
 	persist persistFunc,
 ) (uint64, *shared.MerkleProof, error) {
-	treeCache, tree, err := makeRecoveryProofTree(ctx, treeCfg, merkleHashFunc, nextLeafID, parkedNodes)
+	treeCache, tree, err := makeRecoveryProofTree(ctx, treeCfg, merkleHashFunc, nextLeafID)
 	if err != nil {
 		return 0, nil, err
 	}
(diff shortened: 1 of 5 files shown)


noamnelke (approved):
:shipit:

prover/prover.go
```diff
@@ -181,8 +192,51 @@ func makeRecoveryProofTree(
...
+				return nil, nil, fmt.Errorf("reading parked node in layer %d: %w", layer, err)
+			}
+			logging.FromContext(ctx).
+				Info("recovered parked node", zap.Uint("layer", layer), zap.String("node", fmt.Sprintf("%X", parkedNode)))
+			parkedNodesMap[layer] = parkedNode
+		}
 	}
 
+	// turn parkedNodesMap into a slice ordered by key
+	parkedNodes = parkedNodes[:0]
```
fasmat: Is `parkedNodes` now still necessary as a parameter to the function if it's only used as a backing slice and overwritten anyway?

poszu: No, I will do a cleanup soon. The plan is to remove parked nodes completely from round state.bin and rely solely on implemented recovery.

Btw, I updated PR description.

noamnelke (approved):

fasmat (approved):

PR #349, opened 2023-08-21
Update PoET to Go 1.20

fasmat:


poet.go
```diff
@@ -112,7 +112,7 @@ func poetMain() (err error) {
 		}
 		defer func() {
 			if e := f.Close(); e != nil {
-				err = multierror.Append(err, e).ErrorOrNil()
+				err = errors.Join(err, e)
 			}
```
poszu: How about this:
```go
err = errors.Join(err, f.Close())
```

server/server.go
```diff
@@ -182,9 +179,7 @@ func (s *Server) Start(ctx context.Context) error {
 	// Wait for the server to shut down gracefully
 	<-ctx.Done()
 	grpcServer.GracefulStop()
-	result := multierror.Append(nil, server.Shutdown(context.Background()))
-	result = multierror.Append(result, serverGroup.Wait())
-	return result.ErrorOrNil()
+	return errors.Join(server.Shutdown(context.Background()), serverGroup.Wait())
```
poszu: Is the order of calls to `server.Shutdown()` and `serverGroup.Wait()` guaranteed to be left to right? I don't know about Go, but in C++ the order of evaluation of function arguments is unspecified (https://en.cppreference.com/w/cpp/language/eval_order).

fasmat: Good point, had to look it up because I wasn't sure: https://go.dev/ref/spec#Order_of_evaluation

> ... when evaluating the [operands](https://go.dev/ref/spec#Operands) of an expression, assignment, or [return statement](https://go.dev/ref/spec#Return_statements), all function calls, method calls, and communication operations are evaluated in lexical left-to-right order.

So the code should deterministically execute the two functions in the order `server.Shutdown` then `serverGroup.Wait()`

fasmat: I still updated the code, to avoid confusion. 🙂

poszu (approved):

PR #364, opened 2023-08-31
Split poet into registration and worker services

### LINKED ISSUE #355: Split `Service` into registration and worker
As the first step into _poet v2_, split `Service` into a registration service responsible for registering challenges, scheduling rounds, and serving proofs and into a worker service responsible for round execution, persistence, and proof creation.

### LINKED ISSUE #373: Preserve PoW challenge between restarts
The PoW challenge is updated when a proof is published. However, it is not persisted upon restart. After poet restarts it picks up the "initial pow challenge" from the config.

TODO:
- remove initial pow challenge from the config
- persist the current and previous PoW challenge in DB
- upon poet start:
  - if there is a PoW challenge in DB already - use it as the challenge
  - otherwise, randomly select a new pow challenge

### LINKED ISSUE #380: Avoid flushing 0 pending submits
```
{"L":"INFO","T":"2023-09-04T07:59:56.465Z","N":"/rpc.api.v1.PoetService/Submit","M":"FAILURE","request_id":"d6cac878-bcfa-4c23-bcc1-7e51fee82570","error":"rpc error: code = AlreadyExists desc = conflicting registration"}

2023-09-04T07:59:56.566Z	DEBUG	flushing pending submits	{"num": 0, "round": "3"}
```

poszu:
Closes #355
Closes #373 
Closes #380 

Refactored `Service` by splitting it into 2 entities:
- `Registration`
- `Worker` (still named `Service` in the code)

And two interfaces that they communicate with:
- `WorkerService`
- `RegistrationService`

obraz

## Registration
It's the front-end service that the node talks to. It is responsible for:
- scheduling poet rounds according to round config
- registering new challenges
- persisting rounds' database
- persisting proofs' database
- serving proofs

When a time comes, it atomically closes the open round, builds a membership tree, and opens a new round. The root of the membership tree is passed to the worker service via the `WorkerService` interface.

`Registration` asynchronously waits for proofs coming from the `Worker` and stores them in the DB.

## Worker
It's the backend responsible for executing rounds scheduled by `Registration`. It registers at the `RegistrationService` interface for closed rounds and listens for rounds to execute on the returned channel. When a round is finished, it passes the proof back to the `RegistrationService`.

## `inMemory` transport
It's an entity implementing both `WorkerService` and `RegistrationService` to connect the `Registration` with the `Worker`.  It's very simple and uses 2 channels to pass data between them.


server/state.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/spacemeshos/poet/util"
+)
+
+const stateFilename = "state.bin"
+
+type state struct {
+	PrivKey []byte
+}
+
+func (s *state) save(datadir string) error {
```
fasmat: Both `save` and `loadState` are only called in a single place and I think short enough to be inlined, wdyt? 🙂

util/persistance.go
```diff
@@ -0,0 +1,39 @@
+package util
```
fasmat: I'm not a fan of generic package names like `util`, `shared`, `common`, since they are non-descriptive of what the package contains.

I personally would just inline the two methods (or copy them to server & worker), but if you insist on code de-duplication we could name the package `state` (`state.Persist`, `state.Load`).

util/persistance.go
```diff
@@ -0,0 +1,39 @@
...
+	err = atomic.WriteFile(filename, &w)
+	if err != nil {
+		return fmt.Errorf("writing to disk: %w", err)
+	}
+
+	return nil
+}
+
+func Load(filename string, v any) error {
+	data, err := os.ReadFile(filename) //#nosec G304
```
fasmat: The `G304` warning could be mitigated by only passing the file name instead of the whole path to the function and prepending it here with the path:

```go
	path := filepath.Join(r.datadir, roundStateFileBaseName, filename)
	// or for the server package
	// path := filepath.Join(cfg.DataDir, filename)
	data, err := os.ReadFile(path)
```

util/persistance.go
```diff
@@ -0,0 +1,39 @@
...
+		return fmt.Errorf("loading file: %w", err)
+	}
+
+	_, err = xdr.Unmarshal(bytes.NewReader(data), v)
+	if err != nil {
+		return fmt.Errorf("deserializing: %w", err)
+	}
+
+	return nil
+}
```
fasmat: Using `xdr.NewDecoder` avoids having to read the whole file into memory before decoding it. The same could be done with `xdr.NewEncoder` and then `atomic.ReplaceFile` for storing the state.

```suggestion
func Load(filename string, v any) error {
	f, err := os.Open(filename) //#nosec G304
	if err != nil {
		return fmt.Errorf("loading file: %w", err)
	}

	_, err = xdr.NewDecoder(f).Decode(v)
	if err != nil {
		return fmt.Errorf("deserializing: %w", err)
	}

	return nil
}
```

server/state.go
```diff
@@ -0,0 +1,26 @@
...
+	"path/filepath"
+
+	"github.com/spacemeshos/poet/util"
+)
+
+const stateFilename = "state.bin"
+
+type state struct {
+	PrivKey []byte
+}
```
fasmat: Do we really need XDR to encode the private key? It doesn't add any checksums or does special handling. It just writes the length of the byte slice and then the data, so the file will be:

```
0x00 0x00 0x00 0x20 + bytes of PrivKey
```

I think we could add a small migration to directly write the bytes (or encode them in hex), similar to how we handle the private key for a spacemesh node?

config/round_config/config.go
```diff
@@ -0,0 +1,57 @@
...
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	defaultEpochDuration = 30 * time.Second
+	defaultPhaseShift    = 15 * time.Second
+	defaultCycleGap      = 10 * time.Second
+)
+
+type Config struct {
```
fasmat: Suggestion: instead of having a `round_config` package with `round_config.Config` and `round_config.DefaultConfig()` how about moving this file into `config` and naming those `config.Round` and `config.DefaultRound()`?

For the methods on the config object the same:

- `Config::RoundStart` -> `Round::Start`
- `Config::RoundEnd` -> `Round::End`
- `Config::RoundDuration` -> `Round::Duration`

transport/memory.go
```diff
@@ -0,0 +1,60 @@
...
+
+// inMemory is an in-memory implementation of transport.
+// It allows binding a Registration service with a Worker service
+// in a standalone mode by an in-memory channel.
+type inMemory struct {
+	closedRounds chan service.ClosedRound
+	proofs       chan shared.NIP
+}
+
+func NewInMemory() transport {
```
fasmat, line 24: Would it make sense to use something like `github.com/go-kit/kit` for this? It was built for exactly that use case: enforce service oriented architecture where services can talk over configurable transports with each other (in memory, http, grpc, etc.):

https://github.com/go-kit/kit

Might be worth looking into, here is a simple example on how to use it: https://gokit.io/examples/stringsvc.html and a more complex/complete one: https://github.com/go-kit/examples/tree/master/addsvc

The advantage would be that we only need to focus on implementing the logic of the services themselves and let all the transport, retrying, load balancing, etc. be handled by the library 🙂

transport/memory.go
```diff
@@ -0,0 +1,60 @@
...
+		Epoch:          epoch,
+		MembershipRoot: membershipRoot,
+	}:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (m *inMemory) RegisterForProofs(ctx context.Context) <-chan shared.NIP {
```
fasmat, line 44: The current implementation only allows one client registering for proofs. That's probably OK for in memory, but I'm not sure if this can easily be extended for multiple registered services when using channels (requires larger buffer, or cause signals to be dropped if one client isn't responding)

registration/db.go
```diff
@@ -0,0 +1,146 @@
+package registration
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+
+	xdr "github.com/nullstyle/go-xdr/xdr3"
```
fasmat, line 9: Would it make sense to use `go-scale` instead? Not only should it be faster to decode and encode, the resulting binary data should also be smaller.

registration/db.go
```diff
@@ -0,0 +1,146 @@
...
+	proof := &proofData{}
+	_, err = xdr.Unmarshal(bytes.NewReader(data), proof)
+	if err != nil {
+		return nil, fmt.Errorf("failed to deserialize: %v", err)
+	}
+	return proof, nil
+}
+
+func (db *database) SavePowChallenge(ctx context.Context, challenge []byte) error {
+	trans, err := db.db.OpenTransaction()
```
fasmat: NIT: Usually we denote transactions as `tx`.

```suggestion
	tx, err := db.db.OpenTransaction()
```

registration/doc.go
```diff
@@ -1,12 +1,13 @@
 /*
-Package service provides functionality for creating proofs in dedicated rounds, so that
-they can be shared among arbitrary number of miners, and thus amortize the CPU cost per proof.
+Package registration provides functionality for registering PoET challenges, scheduling rounds, and service proofs.
 
 Each round starts by being open for receiving challenges (byte arrays) from miners.
 Once finalized (according to the configured duration), a hash digest is created from the received challenges,
-and is used as the input for the proof generation.
+and is used as the input for the proof generation. The hash is the root of a merkle tree constructed from registered challenges
+sorted lexigraphically.
```
fasmat:
```suggestion
sorted lexicographically.
```

registration/registration.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+}
+
+func WithConfig(cfg Config) newRegistrationOptionFunc {
+	return func(opts *newRegistrationOptions) {
+		opts.cfg = cfg
+	}
+}
+
+func NewRegistration(
```
fasmat:
```suggestion
func New(
```

registration/registration.go
```diff
@@ -0,0 +1,362 @@
...
+	return r, nil
+}
+
+func (r *Registration) Pubkey() ed25519.PublicKey {
+	return r.privKey.Public().(ed25519.PublicKey)
+}
+
+func (r *Registration) Close() (res error) {
+	return errors.Join(res, r.db.Close(), r.openRound.Close())
+}
```
fasmat:
```suggestion
func (r *Registration) Close() error {
	return errors.Join(r.db.Close(), r.openRound.Close())
}
```

registration/round.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+}
+
+func withMaxSubmitBatchSize(size int) newRoundOptionFunc {
+	return func(o *newRoundOptions) {
+		o.maxSubmitBatchSize = size
+	}
+}
+
+func withFailfIfNotExists() newRoundOptionFunc {
```
fasmat:
```suggestion
func withFailIfNotExists() newRoundOptionFunc {
```

registration/registration.go
```diff
@@ -0,0 +1,362 @@
...
+		select {
+		case <-ctx.Done():
+			logger.Debug("context canceled while waiting for challenge to be submitted")
+			return 0, time.Time{}, ctx.Err()
+		case err := <-done:
+			if err != nil {
+				logger.Warn("challenge registration failed", zap.Error(err))
+				return 0, time.Time{}, err
+			}
+		}
```
fasmat: This seems incorrect. Why is `r.openRound.submit` not just a blocking call that returns an error if it fails?

```suggestion
	err := r.openRound.submit(ctx, nodeID, challenge)
	if err != nil {
		logger.Warn("challenge registration failed", zap.Error(err))
		return 0, time.Time{}, err
	}
```

The `select` on the context while waiting for a possible `error` on `done` in my opinion belongs into the `submit` of the `openRound`?

Calls to `submit` are mutually exclusive anyway, so I see no reason to just make this a blocking call?

registration/round.go
```diff
@@ -0,0 +1,265 @@
...
+	return mtree.Root(), nil
+}
+
+func (r *round) getMembers() (members [][]byte) {
+	iter := r.db.NewIterator(nil, nil)
+	defer iter.Release()
+	for iter.Next() {
+		member := make([]byte, len(iter.Value()))
+		copy(member, iter.Value())
+		members = append(members, member)
```
fasmat: NIT append also copies if used like this and avoids the double call to `iter.Value()`:

```suggestion
		var member []byte
		member = append(member, iter.Value()...)
		members = append(members, member)
```

This is results in a copy as can be seen here: https://go.dev/play/p/orMHwAnr48w

service/round.go
```diff
@@ -1,63 +1,46 @@
...
-	})
-)
+var leavesMetric = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "poet",
+	Subsystem: "round",
+	Name:      "leaves_total",
+	Help:      "Number of generated leaves in a round",
+}, []string{"id"})
+
+// TODO(brozansk): align with `executionState` when migrated to new state structure.
```
fasmat: Do we need an Issue for this?

service/service.go
```diff
@@ -2,308 +2,199 @@ package service
...
+		o.roundCfg = cfg
 	}
 }
 
 // NewService creates a new instance of Poet Service.
 // It should be started with `Service::Run`.
-func NewService(ctx context.Context, cfg *Config, dbdir, datadir string, opts ...OptionFunc) (*Service, error) {
-	options := option{
-		verifier: NewPowVerifier(NewPowParams([]byte(cfg.InitialPowChallenge), cfg.PowDifficulty)),
+func NewService(
```
fasmat: NIT:

```suggestion
func New(
```

server/state.go
poszu: I think the idea was that the state could be extended this way. It doesn't hurt to embed the key into a structure but it makes it more flexible.

registration/db.go
poszu: Yes! However, this needs migration as the DB is already using xdr. Will create a separate issue for this.

registration/registration.go
poszu: What's the benefit?

config/round_config/config.go
poszu: I had to move it into a separate package to avoid cyclic dependency: `package config` depends on `package service` which depends on `package round_config`. :(

transport/memory.go
poszu: I'm not sure, it seems an overkill at the moment. It would make the implementation more complex than it needs to be. I'd reconsider it when additional transport is actually required (when we do the split into separate processes in #365. Wdyt?

transport/memory.go
poszu: Yes, it supports only 1 "consumer" registered. I'd rather not complicate this until there's actually a need for it. It will probably be needed eventually for the distributed k2pow. The k2pow service will be the second consumer listening for new proofs then.

registration/registration.go
poszu: The `round::submit()` is async because I don't want to hold the **shared** (`RLock()`) lock of `openRoundMutex` until the batch is flushed. It's enough for the submit to be batched and the lock can be released (btw, the `openRoundMutex.Unlock()` is missing which is a bug - fixing right now).

When the round is about to be closed, the `openRoundMutex` is obtained **exclusively** (`Lock()`), and the pending batch is flushed (completing all pending `Registration::Submit()`). The round is executed and a new round is opened.

💡 A blocked call to `RWMutex::Lock()` prevents subsequent `RWMutex::RLock()` from acquiring the lock even if it is already held for read by other goroutines, which prevents starvation of the main `Registration` thread that tries to close the open round (https://pkg.go.dev/sync#RWMutex).

util/persistance.go
poszu: Changed to use `os.Open()`. Kept using `xdr.Unmarshal()` as it is the exactly same thing as `xdr.NewDecoder(f).Decode(v)`.

util/persistance.go
poszu: The poet doesn't process any paths coming from the users hence I just silenced it.

util/persistance.go
poszu: Renamed the package to `state`.

service/round.go
poszu: Done.

server/state.go
fasmat: It only becomes more flexible if the file also contains some meta information about the data stored inside of it. Otherwise if we extend `state` in the future, there is no difference between having an object or just bytes.

With my comment I wanted to say we don't need XDR to encode a private key; `go-scale` would do the same thing but only use one byte instead of 4 to encode the length (and be faster in doing so), if we encode in hex it would be less efficient but human readable (the primary reason we encode in hex everywhere else).

registration/registration_test.go
```diff
@@ -165,17 +165,20 @@ func TestPowChallengeRotation(t *testing.T) {
...
 		DoAndReturn(func(_ context.Context, epoch uint, _ []byte) error {
-			proofs <- shared.NIP{
+			select {
+			case proofs <- shared.NIP{
 				MerkleProof: shared.MerkleProof{
 					Root: []byte{1, 2, 3, 4},
 				},
 				Epoch: epoch,
+			}:
+			default:
```
fasmat, line 174: Should there be an `require.Fail()` if the `NIP` couldn't be sent to proofs?

config/round_config/config.go
fasmat: Hm... there seems to be a weird relationship between a few packages:

- both `server` and `rpc` depend on `config`
- `server` depends on `rpc`
- Moving `config` into `server` (imo it belongs there) causes a cyclic dependency

This can be avoided: `rpc` doesn't need to depend on `config`. It only uses 2 fields out of the whole config object and they could be passed explicitly (`PhaseShift` and `CycleGap`).

`config` and `server` both depend on `round_config`, but `registration` as well (which is imported by `server`). This I guess is the reason why you split them into two.

`round_config` could be merged with the other config if the methods defined on `round_config.Config` would be functions defined in the `registration` package (only receiving the relevant fields).

Wdyt?

transport/memory.go
fasmat: In the long run I think it would make it simpler than complex. It would force us to clearly define our DTOs and service internal data structures which leads to better decoupling and easier testing: it would also avoid the cyclic dependency issue between `config`, `round_config` and other packages. If they are defined in `server` they can only be used there and other services would need their own configuration (that might be passed via an API or loaded during startup, depending on the use case).

Yes, within this PR this would be too much, but I feel like it is worth giving a try for #365

registration/db.go
fasmat: Can you reference the issue here so we don't forget it? 🙂

registration/registration.go
fasmat: The first one is basically:

```go
return errors.Join(nil, r.db.Close(), r.openRound.Close())
```

causing 2 extra loop iterations: https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/errors/join.go;l=15

registration/registration.go
fasmat: What I mean is the `Submit` method in `registration.go` could look like this:

```go
func (r *Registration) Submit(
	ctx context.Context,
	challenge, nodeID []byte,
	nonce uint64,
	powParams PowParams,
) (epoch uint, roundEnd time.Time, err error) {
	logger := logging.FromContext(ctx)

	err = r.powVerifiers.VerifyWithParams(challenge, nodeID, nonce, powParams)
	if err != nil {
		logger.Debug("challenge verification failed", zap.Error(err))
		return 0, time.Time{}, err
	}
	logger.Debug("verified challenge", zap.String("node_id", hex.EncodeToString(nodeID)))

	r.openRoundMutex.RLock()
	epoch = r.openRound.epoch
	r.openRoundMutex.RUnlock()
	err = r.openRound.submit(ctx, nodeID, challenge)

	switch {
	case err == nil:
	case errors.Is(err, ErrChallengeAlreadySubmitted):
	case err != nil:
		return 0, time.Time{}, err
	}

	return epoch, r.roundCfg.RoundEnd(r.genesis, epoch), nil
}
```

If we update the `submit` method to:

```go
func (r *round) submit(ctx context.Context, key, challenge []byte) error {
	done, err := r.submitLocked(ctx, key, challenge)
	if err != nil {
		return err
	}

	select {
	case <-ctx.Done():
		return ctx.Err()
	case err := <-done:
		return err
	}
}

func (r *round) submitLocked(ctx context.Context, key, challenge []byte) (<-chan error, error) {
	r.batchMutex.Lock()
	defer r.batchMutex.Unlock()

	if r.members+len(r.pendingSubmits) >= r.maxMembers {
		return nil, ErrMaxMembersReached
	}

	pending, ok := r.pendingSubmits[string(key)]
	if ok {
		switch {
		case bytes.Equal(challenge, pending.challenge):
			return nil, fmt.Errorf("%w: key: %X", ErrChallengeAlreadySubmitted, key)
		default:
			return nil, ErrConflictingRegistration
		}
	}

	if r.batch == nil {
		r.batch = leveldb.MakeBatch(r.maxBatchSize)
		time.AfterFunc(r.flushInterval, r.flushPendingSubmits)
	}

	registered, err := r.db.Get(key, nil)
	switch {
	case errors.Is(err, leveldb.ErrNotFound):

	// OK - challenge is not registered yet.
	case err != nil:
		return nil, fmt.Errorf("failed to check if challenge is registered: %w", err)
	case bytes.Equal(challenge, registered):
		return nil, fmt.Errorf("%w: key: %X", ErrChallengeAlreadySubmitted, key)
	default:
		return nil, ErrConflictingRegistration
	}

	r.batch.Put(key, challenge)
	done := make(chan error, 1)
	r.pendingSubmits[string(key)] = pendingSubmit{
		done,
		challenge,
	}

	if r.batch.Len() >= r.maxBatchSize {
		r.flushPendingSubmitsLocked()
	}
	return done, nil
}
```

`submitLocked` is basically the same as the current `submit` method but this alleviates the caller from having to check 2 different errors (`error` & `done`). Simplifying the call in `registration.go`.

registration/registration_test.go
poszu: IMHO no. Because the mocked `Execute` sends directly to the proofs channel and because the epoch is so short in the test, it can happen that `Execute` is called multiple times by the Registration scheduler and starves the proof-consuming branch. Then the `Execute` might get stuck when the test ends (context is cancelled and the scheduler quits, no longer consuming the proofs). This is not going to happen in normal circumstances. I could add an additional code at the end of `Registration::Run()` to consume all pending proofs to mitigate this and avoid such "hacks" in tests

registration/registration.go
poszu: The `Registration::Submit()` is called concurrently with `Registration::closeRound()`. The proposed code is racy and could lead to submitting to a closed round or writing batch to a closed DB etc.

fasmat: I disagree, if the code wasn't racy before it also isn't racy with the proposed change. Nothing regarding the used mutexes has changed. Calling `submit` does the same thing as before, just that the `done` channel is read from before returning instead of giving the caller that responsibility.

server/state.go
poszu: > It only becomes more flexible if the file also contains some meta information about the data stored inside of it. Otherwise if we extend state in the future, there is no difference between having an object or just bytes.

The meta information could be in the filename ;) i.e `state.bin`, `state_v1.bin` etc.

Anyway, don't get me wrong - I'm not defending it - you asked why is it a struct and that's my guess. This code has already been there and changing it in this PR is out of its scope. Personally, I'd migrate all the state files into JSON.

registration/db.go
poszu: https://github.com/spacemeshos/poet/issues/386

registration/registration.go
poszu: True, I'm dumb and somehow looked at the suggested diff in the inverted direction 😅

config/round_config/config.go
poszu: >Moving config into server (imo it belongs there)

Makes sense :+1:

>config and server both depend on round_config, but registration as well (which is imported by server). This I guess is the reason why you split them into two.

Yes, exactly.

>round_config could be merged with the other config if the methods defined on round_config.Config would be functions defined in the registration package (only receiving the relevant fields).

The `round_config.Config` is not part of the `registration` package because it is used by both `registration` and `service` packages. The round start and end calculations are complex enough that I prefer to have them in a single place rather than inline everywhere and risk making a mistake.

registration/registration.go
poszu: It is racy. The code wasn't racy before (that I know of). The challenge used to be batched with `r.openRoundMutex` Rlocked which prevented closing the round. Now this is not true. Consider:
1. `Registration::Submit(challengeA)` is called, executes until right after Unlock:
    ```go
    r.openRoundMutex.RUnlock()
    // halts here
    err = openRound.submit(ctx, nodeID, challenge) // note: removed `r.`, I assume you meant it this way.
    ```
2. `r.closeRound()` is called. It can take `r.openRoundMutex.Lock()`
3. The root is calculated `root, err := r.openRound.calcMembershipRoot()`. It doesn't contain `challengeA`
5. The `challengeA` submit (1.) continues, writes the challenge into a batch too late as the root is already calculated
6. the submit finishes and falsely returns OK to the user

config/round_config/config.go
poszu: Done

fasmat (approved):

PR #375, opened 2023-09-06
Skip mocks in codecov

poszu:
Don't include mocks in the Codecov report.


fasmat (approved):

fasmat:
Would it make sense to also apply this to `go-spacemesh` and `post`?

PR #391, opened 2023-09-13
Poet Operator Manual.

pigmej:


docs/poet_config_sample.conf
```diff
@@ -0,0 +1,9 @@
+genesis-time=2023-07-14T08:00:00.000Z
+epoch-duration=336h
+phase-shift=240h
+cycle-gap=12h
+jsonlog=true
+pow-difficulty=20
+pow-challenge=A_STRING_USED_FOR_CHALLENGE
```
poszu: It's not required (doesn't exist in fact) since v0.9.2 ;) The challenge is picked randomly on the first boot and persisted in DB.
```suggestion
```

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+
+Proof of Elapsed Time, or PoET, is a cryptographic primitive used to prove that a specific amount of real (clock) time has elapsed.
+
+You can think about PoET like an HTTP web service with a twist.
+
+## PoET important endpoints
+
+* https://POET_URL/v1/pow_params
+    nodes will query that URL to get the POW challenge parameters.
+    it will change every round, safe to cache in all other moments.
```
poszu: It could also change if the poet had been restarted with a different `pow-difficulty`, in which case the cache should be invalidated.

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+* https://POET_URL/v1/info
+    contains the current round id and the current round state. Nodes currently don't actively use that endpoint besides querying and checking if poet is alive. Safe to cache and invalidate when round state changes.
```
poszu, line 13: It contains:
- open round ID
- executing round ID
- poet's public key
- round configuration (phase shift, cycle gap)

source: https://github.com/spacemeshos/poet/blob/4ab25f7a5b9515cebd452a228e95dd72d0ad03ff/rpc/api/v1/api.proto#L81-L88

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+* https://POET_URL/v1/submit
+    nodes will submit their PoET proofs to that endpoint. *Never cache*. POW is used to protect PoET from "random submissions".
```
poszu: The PoW doesn't really protect from submitting garbage.
```suggestion
    nodes will submit their PoET proofs to that endpoint. *Never cache*. The POW serves as a rate-limiting factor and is meant to protect PoET from DOS.
    The higher the value of `pow-difficulty`, the bigger the protection but also more burden on the nodes.
```

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+* https://POET_UR/v1/proofs/{round_id}
+    nodes will query that endpoint to get the PoET proofs for a given round. Once data is there then valid forever. Safe to cache.
+
+PoET uses HTTP to GRPC translation. So publicly facing endpoint is HTTP based but internally it goes via GRPC.
```
poszu:
```suggestion
PoET exposes the above API as:
- REST endpoints via the [GRPC gateway](https://github.com/grpc-ecosystem/grpc-gateway)
- GRPC

The reference node implementation [go-spacemesh](https://github.com/spacemeshos/go-spacemesh) expects the REST API so it's recommended to expose it publicly. Exposing GRPC is not required.
```

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+When the round is open poet is pretty much a web application with people using `info` `pow_params` `submit` endpoints. There is a metric `grpc_server_handled_total` that will give the exact numbers etc more info in (#metrics).
+
+### Config files
+
+Please see [poet.service](./poet.service) and [poet_config_sample.conf](./poet_config_sample.conf).
+
+
+## PoET Round States and the tree
+
+When the poet round is `in-progress` it generates the PoSW - proof of sequential work (essentially merkle tree). It will write the tree to a disk to a directory specified by `--datadir` flag. The per round tree will be saved in that directory in a subdirectory named after the round id.
```
poszu:
```suggestion
When the poet round is `in-progress` it generates the PoSW - proof of sequential work (essentially merkle tree). It will write the tree to a disk to a directory specified by `--datadir` flag. The per round tree will be saved in that directory in a subdirectory named after the round id (`<datadir>/rounds/<round-id>`).
```

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+├── layercache_1.bin
+├── layercache_2.bin
+├── layercache_3.bin
+├── layercache_4.bin
+├── layercache_5.bin
+├── layercache_6.bin
+├── layercache_7.bin
+├── layercache_8.bin
+├── layercache_9.bin
+└── state.bin
```
poszu:
```suggestion
<datadir>
└── rounds
    └── 3
        ├── layercache_0.bin
        ├── layercache_1.bin
        ...
        └── state.bin
```

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Layercache
+
+PoET tries to optimize the cache usage by knowing how much `lps` (leafs per second) is it expected to get. You can specify that with `--lps` flag. To obtain the value the best is to use `poet_bench` tool but please run it for significant amount of time (at least 10 minutes) to get the stable results.
+
+By default poet uses 26 memory layers ( Up to (1 << 26) * 2 - 1 Merkle tree cache nodes (32 bytes each) will be held in-memory), the other layers are stored on disk. If your poet generates 10_000 leafs per second then the expected disk storage for a 13,5days running would be `10_000 * 60 * 60 * 24 * 13,5 = 11_664_000_000` leafs. That means that the produced tree will have `11_664_000_000`. But to produce `11_664_000_000` leafs you need `11_664_000_000 - 1` nodes above them. So we can multiply `11_664_000_000` by 2 to get the total tree size which is: `23_328_000_000`. If you multiply it by `32bytes` (size of one node) then you get `373.248 * 2 GB` which is `746,496 GB` for a PoET service that runs for 13,5 days and generates 10_000 leafs per second. That's an approximation because the tree is not always full. But it's a good enough approximation.
```
poszu: 💡 That's the size of the tree, not the size on the disk as some layers are stored in RAM.

... but on the other hand, 26 layers are roughly 2GiB.

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+
+Layers are written to a disk with buffered file, you can configure the buffer size by using `tree-file-buffer`, the default is `4096` and should be good enough for most cases.
+
+You can also use that logic to estimate needed disk performance and IOps. But HDD without a SSD cache should be more than sufficient in any case.
+
+You *need* to make sure that the data is persisted. The existence of the tree is essential to generate a proof for a given round. If the tree is not there then the proof will be impossible to properly generate.
```
poszu:
```suggestion
You *must* make sure that the data is persisted. The existence of the tree is essential to generate proof for a given round. If the tree is not there then it's not possible to generate the tree. The proof is a Merkle Proof proving selected `T` leaves from the tree and basically contains selected nodes from the tree.
```

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+
+> [!NOTE]  
+> PoET will *not* start generating tree when it will detect that the round should be already in progress but the state of the round is empty. This is to prevent the situation when PoET is doing work for nothing
```
poszu, line 75: That's not entirely true.

It will execute a round without any members. It's not optimal and could be improved. 

But it will NOT create and execute an empty round when it is started in the middle of a round.

docs/poet_operator_manual.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Performance fine tunning
+
+Because PoET is single threaded and CPU bound, it is important to fine tune the CPU and scheduler to get the best performance. On Linux platform we strongly recommend using kernel 6.2 and newer.
```
poszu: Only the worker building the PoSW is single-threaded. The registration is multithreaded.

docs/poet_operator_manual.md
pigmej: Didn't want to document all the endpoints, honestly, more like say what is it and what for than an exact docu.

docs/poet_operator_manual.md
pigmej: Yeah that's why I said "good enough approximation" and 2GiB changes kinda nothing in 800GiB calculation :)

docs/poet_operator_manual.md
pigmej: Yeah but that's what I mean by "empty state of the round" not that it has no members but that it has not run yet etc. Could try to think how to make it more obvious.

docs/poet_operator_manual.md
pigmej: Yes but I focused here about the PoSW, will add clarification.

docs/poet_operator_manual.md
pigmej: Added short clarification.

poszu (approved):

PR #392, opened 2023-09-14
Bump to Go 1.21

### LINKED ISSUE #363: Bump to Go 1.21

poszu:
Closes #363 

Bumped the minimum Go version to 1.21.

🚀 Benchmarks show ~0.7% improvement in PoSW speed

### DIFF
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -3,7 +3,7 @@ package verifier
 import (
 	"bytes"
 	"fmt"
-	"sort"
+	"slices"
 
 	"github.com/spacemeshos/merkle-tree"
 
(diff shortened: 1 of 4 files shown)


verifier/verifier.go
```diff
@@ -60,12 +60,12 @@ func Validate(proof shared.MerkleProof, labelHashFunc func(data []byte) []byte,
...
+	ret := make([]uint64, 0, len(s))
 	for key, value := range s {
 		if value {
 			ret = append(ret, key)
 		}
 	}
-	sort.Slice(ret, func(i, j int) bool { return ret[i] < ret[j] })
+	slices.Sort(ret)
 	return ret
 }
```
fasmat, line 71: The whole function can be written as:

```go
func asSortedSlice(s map[uint64]bool) []uint64 {
	ret := maps.Keys(s)
	slices.Sort(ret)
	return ret
}
```

if you add the `golang.org/x/exp/maps` dependency. Some of the functions in there have already been integrated into the standard library via the `maps` package, and it is likely that the others will eventually be part of the standard library as well 🙂

fasmat (approved):

verifier/verifier.go
poszu: I know, `maps.Keys` could also be used in different places. I didn't want to add a new dep in this PR though.

fasmat:
For dependabot to work with go 1.21 you have to specify a minor version in the `go.mod` file. See also https://github.com/spacemeshos/go-spacemesh/pull/5016

PR #394, opened 2023-09-18
Fix dependabot not working

fasmat:
Dependabot is currently not working for `spacemeshos/poet`. To verify run:

```
go install github.com/dependabot/cli/cmd/dependabot@latest
dependabot update go_modules spacemeshos/poet
```

It will fail with the error:

```
updater | Dependabot encountered '1' error(s) during execution, please check the logs for more details.
updater | +-------------------------------+
updater | |            Errors             |
updater | +-------------------------------+
updater | | dependency_file_not_parseable |
updater | +-------------------------------+
```

The proposed change fixes that by adding a patch version to `go.mod` and updating `ci.yml` to still use the newest 1.21.x when building the source code.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/spacemeshos/poet
 
-go 1.21.0
+go 1.21.1
 
 require (
 	github.com/c0mm4nd/go-ripemd v0.0.0-20200326052756-bd1759ad7d10
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/spacemeshos/poet
 
-go 1.21
+go 1.21.0
 
 require (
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/spacemeshos/poet
 
-go 1.21
+go 1.21.0
```
poszu: Why not?
```suggestion
go 1.21.1
```

poszu (approved):

go.mod
fasmat: I updated the version in `go.mod`.

The version specified here only says what is the **minimum** version required to build the module, the CI should use the newest 1.21.x anyway. `govulncheck` reports what I think is a false positive: It says that `poet.poetMain` indirectly calls `html/template.Template.Execute` which I don't think is true 🤔 but had a bug that was fixed in `go 1.21.1` so I updated it to be sure.

PR #408, opened 2023-09-26
Update protobuf toolset

poszu:
Updated tools related to protobuf and GRPC codegen.


fasmat (approved):
Could you do the same for the `api` repository? 🙂

PR #413, opened 2023-09-28
Remove open and executing round IDs from /Info GRPC endpoint

poszu:
This change removes unused fields in a backward-compatible way by marking them _reserved_. Neither of them is used by the node.


registration/registration.go
```diff
@@ -195,8 +195,11 @@ func (r *Registration) Run(ctx context.Context) error {
...
 
 	// First re-execute the in-progress round if any
-	if err := r.recoverExecution(ctx); err != nil {
-		return fmt.Errorf("recovering execution: %w", err)
+	if r.openRound.epoch > 0 {
+		epoch := uint(int(r.openRound.epoch) - 1)
+		if err := r.recoverExecution(ctx, epoch); err != nil {
+			return fmt.Errorf("recovering execution: %w", err)
+		}
 	}
```
fasmat: You are already checking that `epoch > 0` so the cast shouldn't be necessary. Even if `epoch` would be 0 with or without the cast the result would be `math.MaxUint`

```suggestion
	if r.openRound.epoch > 0 {
		if err := r.recoverExecution(ctx, r.openRound.epoch-1); err != nil {
			return fmt.Errorf("recovering execution: %w", err)
		}
	}
```

fasmat (approved):

PR #417, opened 2023-10-05
Allow running registration service only

### LINKED ISSUE #415: Allow running registration service only
Add a flag to run the registration service only and don't generate the POET proof.

poszu:
Closes #415 

A new `--disable-worker` flag allows for running the registration service only.

### DIFF
--- a/registration/registration_test.go
+++ b/registration/registration_test.go
@@ -158,6 +158,8 @@ func TestWorkingWithoutWorkerService(t *testing.T) {
 	defer cancel()
 	eg.Go(func() error { return reg.Run(ctx) })
 
+	// Verify that registration keeps opening rounds even without the worker service.
+	// Only check if the round number is incremented to not rely on the exact timing.
 	for i := 0; i < 3; i++ {
 		round := reg.OpenRound()
 		require.Eventually(t, func() bool { return reg.OpenRound() > round }, time.Second, time.Millisecond*10)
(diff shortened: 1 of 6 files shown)


registration/registration_test.go
```diff
@@ -137,6 +140,32 @@ func TestOpeningRounds(t *testing.T) {
...
+	t.Cleanup(func() { require.NoError(t, reg.Close()) })
+
+	var eg errgroup.Group
+	ctx, cancel := context.WithCancel(logging.NewContext(context.Background(), zaptest.NewLogger(t)))
+	defer cancel()
+	eg.Go(func() error { return reg.Run(ctx) })
+
+	for i := 0; i < 3; i++ {
+		round := reg.OpenRound()
+		require.Eventually(t, func() bool { return reg.OpenRound() > round }, time.Second, time.Millisecond*10)
```
fasmat, line 165: Would it make sense for this test assert to just assert that `reg.OpenRound()` returns round 0, 1, 2?

For this the `EpochDuration` probably needs to be a bit longer and/or the interval for `require.Eventually` needs to be smaller.

poszu: That's exactly why I didn't want to check if open round increments one by one - to avoid flaky tests. To test the functionality, it just needs to make sure that registration is not stuck (for example on writing to a full channel).

fasmat: OK, but this won't catch other issues like skipped rounds or if rounds have different durations. Maybe a comment clarifying why it is checked like this would help?

fasmat (approved):

PR #422, opened 2023-10-10
Generate empty proofs in registration-only mode

poszu:
Currently, a poet running in registration-only mode (`--disable-worker`) doesn't generate any proofs. It is problematic/suboptimal as a node expects proof from a poet it submitted to and will retry.

After this change, the poet running in registration-only mode will generate a proof with an empty merkle tree (0 leaves) etc. but with a populated members array. Such proof is invalid and will be rejected by nodes, but they will not retry.

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -103,7 +103,6 @@ func TestSubmitSignatureVerification(t *testing.T) {
 	cfg.PoetDir = t.TempDir()
 	cfg.RawRPCListener = randomHost
 	cfg.RawRESTListener = randomHost
-	cfg.Registration.PowDifficulty = 0
 
 	srv, client := spawnPoet(ctx, t, *cfg)
 	t.Cleanup(func() { assert.NoError(t, srv.Close()) })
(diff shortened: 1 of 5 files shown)


server/server_test.go
```diff
@@ -586,3 +582,75 @@ func TestLoadSubmits(t *testing.T) {
...
+	req.Zero(proof.Proof.Leaves)
+	req.Len(proof.Proof.Members, 1)
+	req.Contains(proof.Proof.Members, challenge)
+	cancel()
+
+	merkleProof := shared.MerkleProof{
+		Root:         proof.Proof.Proof.Root,
+		ProvenLeaves: proof.Proof.Proof.ProvenLeaves,
+		ProofNodes:   proof.Proof.Proof.ProofNodes,
+	}
```
fasmat, line 647: NIT: `proof.Proof.Proof.X` seems quite redundant 😅

fasmat (approved):

server/server_test.go
poszu: yes... could be `response.Proof.Proof.X` :shrug:

PR #429, opened 2023-10-30
Support certificates for registeration

### LINKED ISSUE #427: add support for certifier service in poet

poszu:
Closes #427

There is now an alternative way to PoW of verifying if a node is allowed to register in the open round by certificate verification.

It's considered unstable now and the poet supports both ways until it is stabilized.

If a certifier is configured (an URL + a pubkey), the poet will first try to verify the certificate. If it's not provided (for example a node that didn't update tries to register) - fallback to verifying PoW. Hence a poet that wants to support certificates should configure PoW as well until the certificate way is fully stabilized.

It returns _unauthorized_ (HTTP 401) if the certifier is configured but the certificate is not valid, allowing the node the re-certify and retry registration with a new certificate.


poszu:
bors try

registration/config.go
```diff
@@ -11,9 +16,39 @@ func DefaultConfig() Config {
 }
 
 type Config struct {
-	PowDifficulty uint `long:"pow-difficulty" description:"PoW difficulty (in the number of leading zero bits)"`
+	// FIXME: remove depreacated PoW
```
fasmat:
```suggestion
	// FIXME: remove deprecated PoW
```

release/proto/openapiv2/rpc/api/v1/api.swagger.json
```diff
@@ -127,6 +127,28 @@
     }
   },
   "definitions": {
```
fasmat, line 129: I don't think we need the openAPI definition of the poet API. It is usually only used to host a Swagger UI (Example: https://petstore.swagger.io/). We could remove it from `buf.gen.yaml` 🙂

rpc/api/v1/api.proto
```diff
@@ -59,8 +59,10 @@ message PowParamsResponse {
 
 message SubmitRequest {
     // Proof of Work nonce
+    // deprecated - use certificate instead
```
fasmat, line 62: Would it make sense to instead have a dedicated endpoint for "submit with certificate"? The original one would take `nonce` and `pow_params` and the new one the certificate.

This way when we retire the old handler we can just return 404 or a similar error when it is called.

fasmat (approved):

release/proto/openapiv2/rpc/api/v1/api.swagger.json
poszu: Opened https://github.com/spacemeshos/poet/issues/434 to do it.

rpc/api/v1/api.proto
poszu: That would work for sure, but I think it would complicate the client and CF setup w/o real benefit.

fasmat: CF?

poszu: cloudflare

PR #435, opened 2023-11-24
Allow providing the key via env

poszu:
Added an option to provide the private key via env variable named `POET_PRIVATE_KEY`. It's meant mostly for testing (systests).

### DIFF
--- a/server/state.go
+++ b/server/state.go
@@ -1,0 +1,68 @@
+package server
+
+import (
+	"bytes"
+	"context"
+	"crypto/ed25519"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"go.uber.org/zap"
+
+	"github.com/spacemeshos/poet/logging"
+	"github.com/spacemeshos/poet/state"
+)
+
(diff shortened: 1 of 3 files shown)


server/state.go
```diff
@@ -0,0 +1,68 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("generating key: %w", err)
+		}
+		s.PrivKey = privateKey
+
+		log.Info("generated new keys", zap.Binary("public key", pubKey))
+	case errors.Is(err, os.ErrNotExist) && key != nil:
+		s.PrivKey = key
+	case err == nil && key != nil && !bytes.Equal(key, s.PrivKey):
+		return nil, fmt.Errorf("private key mismatch. env: %s != %s: %s", key, stateFilename, s.PrivKey)
```
fasmat: NIT: avoid printing secrets / keys to stdout

```suggestion
		return nil, fmt.Errorf("private key mismatch. env != %s", stateFilename)
```

fasmat (approved):

server/state.go
poszu: Gooood catch, thanks!

PR #450, opened 2024-01-09
Run on macos-13-xlarge instead of self hosted

fasmat:
Execute unit tests on GH runner instead of self hosted M1


fasmat:
We stick with self hosted for now

PR #458, opened 2024-02-13
Fix scheduling round batch flush

poszu:
Submitting the first challenge in a batch schedules a timed flush. However, if the batch later reaches its maximum size it's automatically flushed. In this situation, the timed flush is not canceled, and flushes the batch after the interval. In the condition of a high submit load, the batches are quickly filled up and manually flushed. **Multiple timed flushes are created and pending simultaneously.** While not causing any inconsistencies, it's inefficient.

This PR changes the logic to cancel the pending timed flush when the batch is flushed manually.


registration/round.go
```diff
@@ -180,14 +181,18 @@ func (r *round) submit(ctx context.Context, key, challenge []byte) (<-chan error
...
 		logging.FromContext(ctx).Debug(
 			"scheduling flush of pending submits",
 			zap.Uint("round", r.epoch),
 			zap.Duration("interval", r.flushInterval),
 		)
-		time.AfterFunc(r.flushInterval, r.flushPendingSubmits)
+		r.pendingFlush = time.AfterFunc(r.flushInterval, func() {
+			r.batchMutex.Lock()
+			defer r.batchMutex.Unlock()
+			r.flushPendingSubmitsLocked(true)
```
fasmat: The boolean parameter isn't really needed or is it?

```suggestion
			r.batchMutex.Lock()
			defer r.batchMutex.Unlock()
			if r.pendingFlush != nil {
				r.flushPendingSubmitsLocked()
			}
```

You can even put this into its own method to make calling it from the test easier 🙂 

```go
func (r *round) timedFlushPendingSubmits() {
	r.batchMutex.Lock()
	defer r.batchMutex.Unlock()
	if r.pendingFlush != nil {
		r.flushPendingSubmitsLocked()
	}
}
```

registration/round.go
```diff
@@ -199,7 +200,19 @@ func (r *round) flushPendingSubmits() {
 	r.flushPendingSubmitsLocked()
 }
 
+func (r *round) timedFlushPendingSubmits() {
+	r.batchMutex.Lock()
+	defer r.batchMutex.Unlock()
+	if r.pendingFlush != nil {
+		r.flushPendingSubmitsLocked()
+	}
```
fasmat, line 208: This `if` is now always true I think?

fasmat (approved):

registration/round.go
poszu: No, it's not. It won't be true when the timer fires while another goroutine is in `flushPendingSubmitsLocked()`. Then, `timedFlushPendingSubmits()` waits on `r.batchMutex.Lock()`, while the other goroutine clears the `r.pendingFlush`.

PR #471, opened 2024-03-21
Support for certificates with expiration date

poszu:
Support for certificates with an optional expiration date. A certificate is now a scale-encoded combination of the node's ID and (optional) expiration time. If the expiration time is `nil`, the certificate does not expire.

The function to verify a certificate is exported in `shared` pkg for the node to verify the certs it gets from a certifier.


registration/registration.go
```diff
@@ -323,17 +323,22 @@ func (r *Registration) Submit(
...
 	// Verify if the node is allowed to register.
 	// Support both a certificate and a PoW while
 	// the certificate path is being stabilized.
 	if r.cfg.Certifier != nil && certificate != nil {
-		if !ed25519.Verify(r.cfg.Certifier.PubKey.Bytes(), nodeID, certificate) {
+		_, err := shared.VerifyCertificate(certificate, r.cfg.Certifier.PubKey.Bytes(), nodeID)
+		switch {
+		case errors.Is(err, shared.ErrCertExpired):
+			registerWithCertMetric.WithLabelValues("expired").Inc()
+			return 0, time.Time{}, fmt.Errorf("%w: %v", ErrInvalidCertificate, err)
```
fasmat: You could use `errors.Join` here:

```suggestion
			return 0, time.Time{}, errors.Join(ErrInvalidCertificate, err)
```

registration/registration.go
```diff
@@ -323,17 +323,22 @@ func (r *Registration) Submit(
...
+		case err != nil:
 			registerWithCertMetric.WithLabelValues("invalid").Inc()
-			return 0, time.Time{}, ErrInvalidCertificate
+			return 0, time.Time{}, fmt.Errorf("%w: %v", ErrInvalidCertificate, err)
```
fasmat: Same as above:

```suggestion
			return 0, time.Time{}, errors.Join(ErrInvalidCertificate, err)
```

fasmat (approved):

PR #514, opened 2024-08-31
Trusted pubkeys 

### LINKED ISSUE #501: Add support for accepting additional trusted certifier public keys
The poet accepts registrations with certificates. A certificate is signed using a private key by an entity called *certifier*. The supported certifier is configured in the config: https://github.com/spacemeshos/poet/blob/7a0c2b96c1312b04bdbe59c1e3b1bca6a8b05f7e/registration/config.go#L45-L48
Currently, the poet accepts only 1 certifier public key along with the certifier URL. The URL exposed on the /v1/info endpoint so that the nodes know where they can obtain the certificate. See PR #429  for more details.

This task is to add support for additional trusted public keys (but no associated certifier URLs). 

The additional keys should be stored in a configured directory, and there should be an grpc API call to reread the contents and load the certs from the directory (ignoring any previously configured).

#### Verifying cert signature against multiple keys

Poet needs to match somehow the certificate to on of the configured public keys to verify its signature. One possibility is to pass a hint (e.g. a 4-B prefix of the key) in the `/Submit` payload. The poet could then find the complete key in the list of known keys. In the rare occasion of a conflict (2.32e-10 probability for 2 keys), it could emit a warning on startup and fall back to trying all keys matching the prefix. Alternatively, it could reject such config.

Note: The client (go-spacemesh) must be updated to pass the pubkey hint).

ConvallariaMaj:
Closes https://github.com/spacemeshos/poet/issues/501 

PR includes:
- reading trusted keys from dir (each pubkey must be in separate file)
- API to invoke reloading trusted keys
- verification certs to handle changes above
- new tests

Connected with PR:
https://github.com/spacemeshos/go-spacemesh/pull/6313


server/server.go
```diff
@@ -180,6 +180,8 @@ func (s *Server) Start(ctx context.Context) error {
 	)
 
 	api.RegisterPoetServiceServer(grpcServer, rpcServer)
+	api.RegisterConfigurationServiceServer(grpcServer, rpcServer)
```
poszu: The configuration service should probably not be exposed on the same, publicly available port as the public API.

shared/cert.go
```diff
@@ -79,19 +85,39 @@ func EncodeCert(c *Cert) ([]byte, error) {
...
-	if !bytes.Equal(decoded.Pubkey, nodeID) {
-		return nil, errors.New("pubkey mismatch")
-	}
-	if decoded.Expiration != nil && decoded.Expiration.Before(time.Now()) {
-		return nil, fmt.Errorf("%w at %v", ErrCertExpired, decoded.Expiration)
+func VerifyCertificate(
+	certificate *OpaqueCert,
+	matchingKeys [][]byte,
+	nodeID []byte,
+) (*Cert, error) {
```
poszu: I think it would be more ergonomic to leave this function as it was (take 1 certificate) and have the user call it many times, for each certificate.

registration/registration.go
```diff
@@ -168,6 +177,49 @@ func (r *Registration) CertifierInfo() *CertifierConfig {
...
 
+func (r *Registration) LoadTrustedPublicKeys() error {
+	if r.cfg.Certifier == nil {
+		return ErrCertificationIsNotSupported
+	}
+
+	r.trustedKeysMtx.Lock()
+	defer r.trustedKeysMtx.Unlock()
+
+	r.trustedPublicKeys = [][]byte{}
```
poszu: Should `trustedPublicKeys` be reset immediately? If the code below fails for some reason, the poet would be left w/o the previous keys. How about swapping the configured keys after reading the directory?

registration/registration.go
```diff
@@ -168,6 +177,49 @@ func (r *Registration) CertifierInfo() *CertifierConfig {
...
+		if err = key.UnmarshalFlag(string(keyData)); err != nil {
+			return err
+		}
+
+		if len(key.Bytes()) != ed25519.PublicKeySize {
+			return ErrInvalidPublicKey
+		}
+
+		r.trustedPublicKeys = append(r.trustedPublicKeys, key.Bytes())
+		return nil
```
poszu, line 229: NIT: How about adding an INFO log saying that the new keys were successfully loaded?

server/server_test.go
```diff
@@ -274,6 +276,121 @@ func TestSubmitCertificateVerification(t *testing.T) {
...
+	return path
+}
+
+// Test submitting a challenge followed by proof generation and getting the proof via GRPC.
+func TestLoadTrustedKeysAndSubmit(t *testing.T) {
+	const keysNum = 2
+
+	// create test files
+	dir := t.TempDir()
+	t.Cleanup(func() { os.RemoveAll(dir) })
```
poszu: The temporary directory is already removed automatically: https://pkg.go.dev/testing#T.TempDir

server/server_test.go
```diff
@@ -274,6 +276,121 @@ func TestSubmitCertificateVerification(t *testing.T) {
...
+func createTestKeyFile(t *testing.T, dir, name string, data []byte) string {
+	path := filepath.Join(dir, name)
+	if err := os.WriteFile(path, data, 0o644); err != nil {
+		t.Fatalf("failed to write key file: %v", err)
+	}
+	return path
+}
+
+// Test submitting a challenge followed by proof generation and getting the proof via GRPC.
+func TestLoadTrustedKeysAndSubmit(t *testing.T) {
```
poszu, line 283: It would be great to also add a test to verify if previously configured trusted keys are forgotten after their files were removed and keys were reloaded.

registration/registration_test.go
```diff
@@ -349,8 +358,51 @@ func Test_CheckCertificate(t *testing.T) {
...
+		pubKeyHint := pub[:shared.CertPubkeyHintSize]
+
 		powVerifier := mocks.NewMockPowVerifier(gomock.NewController(t))
 
+		const keysNum = 3
+
+		dir := t.TempDir()
+		t.Cleanup(func() {
+			os.RemoveAll(dir)
+		})
```
fasmat: `t.TempDir()` already adds a `t.Cleanup` `func` that deletes the created directory after the test ends

```suggestion
		dir := t.TempDir()
```

server/server_test.go
```diff
@@ -274,6 +276,121 @@ func TestSubmitCertificateVerification(t *testing.T) {
 	require.NoError(t, eg.Wait())
 }
 
+func createTestKeyFile(t *testing.T, dir, name string, data []byte) string {
+	path := filepath.Join(dir, name)
+	if err := os.WriteFile(path, data, 0o644); err != nil {
+		t.Fatalf("failed to write key file: %v", err)
+	}
```
fasmat: Also this function can be inlined 🙂 

```suggestion
	require.NoError(t, os.WriteFile(path, data, 0o644))
```

server/server_test.go
```diff
@@ -274,6 +276,121 @@ func TestSubmitCertificateVerification(t *testing.T) {
...
+
+	var (
+		trustedKeyCert *shared.OpaqueCert
+		trustedKey     []byte
+	)
+
+	challenge := []byte("challenge")
+
+	// User credentials
+	userPubKey, userPrivKey, err := ed25519.GenerateKey(rand.Reader)
```
fasmat: NIT: passing `nil` causes `GenerateKey` to use a cryptographically secure source for key generation. One only needs to pass a non-`nil` parameter if they want to ensure that specific keys are generated.

```suggestion
	userPubKey, userPrivKey, err := ed25519.GenerateKey(nil)
```

server/server_test.go
```diff
@@ -274,6 +276,121 @@ func TestSubmitCertificateVerification(t *testing.T) {
...
+			data, err := shared.EncodeCert(&shared.Cert{Pubkey: userPubKey, Expiration: &expiration})
+			require.NoError(t, err)
+
+			trustedKeyCert = &shared.OpaqueCert{
+				Data:      data,
+				Signature: ed25519.Sign(private, data),
+			}
+			trustedKey = pubKey
+		}
+	}
```
fasmat: NIT: This can be simplified:

```suggestion
	// generate keys
	for i := range keysNum - 1 {
		pubKey, _, err := ed25519.GenerateKey(nil)
		require.NoError(t, err)

		path := filepath.Join(dir, fmt.Sprintf("valid_key_%d.key", i))
		require.NoError(t, os.WriteFile(path, []byte(base64.StdEncoding.EncodeToString(pubKey)), 0o644))
	}
	// generate key and certificate
	trustedKey, private, err := ed25519.GenerateKey(nil)
	require.NoError(t, err)

	path := filepath.Join(dir, "valid_key.key")
	require.NoError(t, os.WriteFile(path, []byte(base64.StdEncoding.EncodeToString(trustedKey)), 0o644))
	expiration := time.Now().Add(time.Hour)
	data, err := shared.EncodeCert(&shared.Cert{Pubkey: userPubKey, Expiration: &expiration})
	require.NoError(t, err)

	trustedKeyCert = &shared.OpaqueCert{
		Data:      data,
		Signature: ed25519.Sign(private, data),
	}
```

server/server_test.go
```diff
@@ -274,6 +276,121 @@ func TestSubmitCertificateVerification(t *testing.T) {
...
+
+	t.Parallel()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	ctx = logging.NewContext(ctx, zaptest.NewLogger(t))
```
fasmat: NIT: `t.Parallel()` is best called early in the test 🙂

server/server_test.go
```diff
@@ -274,6 +276,121 @@ func TestSubmitCertificateVerification(t *testing.T) {
...
+
+	certifierPubKey, _, err := ed25519.GenerateKey(rand.Reader)
+	require.NoError(t, err)
```
fasmat: NIT: again this can be `nil`

```suggestion
	certifierPubKey, _, err := ed25519.GenerateKey(nil)
	require.NoError(t, err)
```

server/server_test.go
```diff
@@ -274,6 +276,121 @@ func TestSubmitCertificateVerification(t *testing.T) {
...
+	)
+	require.NoError(t, err)
+	t.Cleanup(func() { assert.NoError(t, conn.Close()) })
+
+	configClient := api.NewConfigurationServiceClient(conn)
+
+	var eg errgroup.Group
+	eg.Go(func() error {
+		return srv.Start(ctx)
+	})
```
fasmat, line 319: We should wait for the server to exit before ending the test:

```suggestion
	var eg errgroup.Group
	eg.Go(func() error {
		return srv.Start(ctx)
	})
	t.Cleanup(func() { assert.NoError(t, eg.Wait()) })
```

registration/registration.go
```diff
@@ -168,6 +177,51 @@ func (r *Registration) CertifierInfo() *CertifierConfig {
...
+		if err = key.UnmarshalFlag(string(keyData)); err != nil {
+			return err
+		}
+
+		if len(key.Bytes()) != ed25519.PublicKeySize {
+			return ErrInvalidPublicKey
+		}
+
+		loadedKeys = append(loadedKeys, key.Bytes())
+		return nil
```
fasmat: `Base64Enc` exists to allow to pass base64 encoded command line parameters to PoET. Reading base64 encoded keys from a file can be done with standard library tools only:

```suggestion
		f, err := os.Open(path)
		if err != nil {
			return err
		}

		dec := base64.NewDecoder(base64.StdEncoding, f)
		key, err := io.ReadAll(dec)
		if err != nil {
			return ErrInvalidPublicKey
		}
		if len(key) != ed25519.PublicKeySize {
			return ErrInvalidPublicKey
		}
		loadedKeys = append(loadedKeys, key)
		return nil
```

This also allocates memory for the key only once (when reading it from the file) while the previous approach allocates memory for the same key 2 times: once for the string representation and once for the raw bytes.

registration/registration.go
```diff
@@ -168,6 +177,51 @@ func (r *Registration) CertifierInfo() *CertifierConfig {
...
+	})
+
+	r.trustedKeysMtx.Lock()
+	defer r.trustedKeysMtx.Unlock()
```
fasmat: I feel like this should be at the head of the method. If multiple calls to `LoadKeys` happen we only want one routine to read the disk at a time.

registration/registration.go
```diff
@@ -317,9 +371,46 @@ func (r *Registration) newRoundOpts() []newRoundOptionFunc {
...
+				matchingKeys = append(matchingKeys, certKey)
+			}
+		}
+
+		r.trustedKeysMtx.RUnlock()
+
+		if len(matchingKeys) == 0 {
+			return ErrNoMatchingCertPublicKeys
+		}
+	}
```
fasmat: If it can be avoided, try to write `if`s without `else`. The `append`s in the original code can be quite expensive, since appending data to a slice usually requires growing (and thereby copying) the slice first, also `certKeys` would then point to the same memory address as `r.trustedPublicKeys` which can lead to future errors if someone tries to modify `certKeys` and isn't aware that this isn't save to do.

My suggestion should reduce the amount copied data to a minimum and gets rid of the `else` block:

```suggestion
	matchingKeys := [][]byte{r.cfg.Certifier.PubKey.Bytes()}

	if len(certPubKeyHint) != 0 {
		r.trustedKeysMtx.RLock()
		matchingKeys = make([][]byte, 0)

		for _, certKey := range r.trustedPublicKeys {
			l := min(len(certKey), len(certPubKeyHint))
			if bytes.Equal(certKey[:l], certPubKeyHint) {
				matchingKeys = append(matchingKeys, certKey)
			}
		}
		l := min(len(r.cfg.Certifier.PubKey.Bytes()), len(certPubKeyHint))
		if bytes.Equal(r.cfg.Certifier.PubKey.Bytes()[:l], certPubKeyHint) {
			matchingKeys = append(matchingKeys, r.cfg.Certifier.PubKey.Bytes())
		}

		r.trustedKeysMtx.RUnlock()
		if len(matchingKeys) == 0 {
			return ErrNoMatchingCertPublicKeys
		}
	}
```

registration/registration.go
```diff
@@ -317,9 +371,46 @@ func (r *Registration) newRoundOpts() []newRoundOptionFunc {
...
+	var certErr error
+	for _, key := range matchingKeys {
+		_, err := shared.VerifyCertificate(certificate, key, nodeID)
+		if err == nil {
+			// cert verified successfully
+			break
+		}
+		certErr = err
+	}
+	return certErr
```
fasmat: I'm not sure this code is correct. If multiple `matchingKeys` exist and the first fails to verify, but the second does then this method will return an error? How about this instead:

```suggestion
	for _, key := range matchingKeys {
		_, err := shared.VerifyCertificate(certificate, key, nodeID)
		if err == nil {
			return nil
		}
	}
	return errors.New("failed to verify certificate")
```

registration/registration_test.go
```diff
@@ -349,8 +358,51 @@ func Test_CheckCertificate(t *testing.T) {
...
+				require.NoError(t, err)
+
+				trustedKeyCert = &shared.OpaqueCert{
+					Data:      data,
+					Signature: ed25519.Sign(private, data),
+				}
+
+				trustedKey = encodedPubkey
+			}
+		}
```
fasmat: NIT: Can be simplified

```suggestion
		// generate keys
		for i := range keysNum - 1 {
			pubKey, _, err := ed25519.GenerateKey(nil)
			require.NoError(t, err)

			path := filepath.Join(dir, fmt.Sprintf("valid_key_%d.key", i))
			require.NoError(t, os.WriteFile(path, []byte(base64.StdEncoding.EncodeToString(pubKey)), 0o644))
		}
		// generate key and certificate
		trustedKey, private, err = ed25519.GenerateKey(nil)
		require.NoError(t, err)

		path := filepath.Join(dir, "valid_key.key")
		require.NoError(t, os.WriteFile(path, []byte(base64.StdEncoding.EncodeToString(trustedKey)), 0o644))
		expiration = time.Now().Add(time.Hour)
		data, err = shared.EncodeCert(&shared.Cert{Pubkey: nodeID, Expiration: &expiration})
		require.NoError(t, err)

		trustedKeyCert = &shared.OpaqueCert{
			Data:      data,
			Signature: ed25519.Sign(private, data),
		}
```

registration/registration_test.go
```diff
@@ -393,17 +446,70 @@ func Test_CheckCertificate(t *testing.T) {
...
+			_, _, err = r.Submit(
+				context.Background(), challenge,
+				pubKeyHint, nodeID, 0, r.PowParams(), defaultValidCert, time.Time{})
+			require.NoError(t, err)
+
+			key := registration.Base64Enc{}
+			err = key.UnmarshalFlag(string(trustedKey))
+			require.NoError(t, err)
+
+			trustedKeyHint := key.Bytes()[:shared.CertPubkeyHintSize]
```
fasmat: With my suggestion above this just becomes:

```suggestion
			trustedKeyHint := trustedKey[:shared.CertPubkeyHintSize]
```

This test also depends on the execution of another subtest ("certification check enabled"). If the other test is executed first, it will read `valid_key.key` from disk. If the other test is not executed first, that file won't exist and no keys are read from disk.

registration/registration_test.go
```diff
@@ -415,8 +521,101 @@ func Test_CheckCertificate(t *testing.T) {
...
+
+		t.Run("load invalid public keys", func(t *testing.T) {
+			_, priv, err := ed25519.GenerateKey(nil)
+			require.NoError(t, err)
+
+			createTestKeyFile(t, dir, "invalid_key.key", []byte(base64.StdEncoding.EncodeToString(priv)))
+
+			err = r.LoadTrustedPublicKeys()
+			require.ErrorIs(t, err, registration.ErrInvalidPublicKey)
+		})
```
fasmat: NIT:

```suggestion
		t.Run("load invalid public keys", func(t *testing.T) {
			path := filepath.Join(dir, "invalid_key.key")
			require.NoError(t, os.WriteFile(path, []byte("invalid"), 0o644))

			err = r.LoadTrustedPublicKeys()
			require.ErrorIs(t, err, registration.ErrInvalidPublicKey)
		})
```

registration/registration_test.go
```diff
@@ -415,8 +521,101 @@ func Test_CheckCertificate(t *testing.T) {
...
+		EpochDuration: time.Hour,
+		PhaseShift:    time.Second,
+	}
+
+	workerSvc := mocks.NewMockWorkerService(gomock.NewController(t))
+
+	const keysNum = 3
+
+	dir := t.TempDir()
+	t.Cleanup(func() { os.RemoveAll(dir) })
```
fasmat: NIT: cleanup happens automatically 🙂 

```suggestion
	dir := t.TempDir()
```

registration/registration_test.go
```diff
@@ -415,8 +521,101 @@ func Test_CheckCertificate(t *testing.T) {
...
+
+	for i := 0; i < keysNum; i++ {
+		validKey, _, err := generateTestKeyPair()
+		require.NoError(t, err)
+
+		createTestKeyFile(t, dir, fmt.Sprintf("valid_key_%d.key", i), validKey)
+	}
```
fasmat: NIT: these helper functions aren't needed 🙂 

```suggestion
	for i := range keysNum {
		pubKey, _, err := ed25519.GenerateKey(nil)
		require.NoError(t, err)

		path := filepath.Join(dir, fmt.Sprintf("valid_key_%d.key", i))
		require.NoError(t, os.WriteFile(path, []byte(base64.StdEncoding.EncodeToString(pubKey)), 0o644))
	}
```

registration/registration_test.go
```diff
@@ -415,8 +521,101 @@ func Test_CheckCertificate(t *testing.T) {
...
+
+	certPubKey, _, err := generateTestKeyPair()
+	require.NoError(t, err)
```
fasmat: NIT: same here

```suggestion
	pubKey, _, err := ed25519.GenerateKey(nil)
	require.NoError(t, err)
```

The one use of `certPubKey` below then needs to be replaced with `[]byte(base64.StdEncoding.EncodeToString(pubKey))`

fasmat (changes requested):
Tests share a lot of (setup) code. Some tests even start failing if executed individually or in different orders.

A quick fix for this is to just move the shared setup code into a function:

```go
func Test_Test1(t *testing.T) {
    setup := func(t *testing.T) {
        // setup code
    }
    t.Run("subtest1", func(t *testing.T) {
         setup(t)
         // execute test
    })
```

This ensures that all tests setup their environment individually and that they don't leak into other tests.

registration/registration.go
poszu: IMHO, a map like `matchingKeys map[hint][][]byte` could be built when the keys are loaded instead of on every request.

fasmat: Sure, `trustedKeys` can even be replaced with such a mapping. Loading would just put the keys in it and the mutex would protect access to the map.

poszu: Done

registration/registration.go
poszu: I'm not sure, reading the dir in parallel is maybe inefficient from the GRPC request point of view, but it is not wrong. The mutex guards access to the keys map. On the other hand, if the critical section included reading the directory tree, it would be significantly longer which would harm efficiency in the code that we really care about - the **submits**.

poszu:
> Tests share a lot of (setup) code. Some tests even start failing if executed individually or in different orders.
> 
> A quick fix for this is to just move the shared setup code into a function:
> 
> ```go
> func Test_Test1(t *testing.T) {
>     setup := func(t *testing.T) {
>         // setup code
>     }
>     t.Run("subtest1", func(t *testing.T) {
>          setup(t)
>          // execute test
>     })
> ```
> 
> This ensures that all tests setup their environment individually and that they don't leak into other tests.

Which tests depend on the order they are executed in?

server/server.go
poszu: Done

server/server_test.go
poszu: Done

registration/registration.go
fasmat: Yeah I feel like that the whole method should be mutually exclusive, not just the update of the field to avoid that issue. Wdyt?

fasmat:
> Which tests depend on the order they are executed in?

`"valid certificate (with loaded trusted keys)"` if executed individually does not read key files from disk:

```shell
go test -run ^Test_CheckCertificate$/^valid_certificate_\(with_loaded_trusted_keys\)$ github.com/spacemeshos/poet/registration
```

but it actually looks like if executed like this the test isn't run at all. Maybe that's because the tests are double nested?

registration/registration.go
```diff
@@ -317,12 +387,51 @@ func (r *Registration) newRoundOpts() []newRoundOptionFunc {
...
+
+	var certErrors error
+	for _, key := range matchingKeys {
+		_, err := shared.VerifyCertificate(certificate, key, nodeID)
+		if err == nil {
+			return nil
+		}
+		certErrors = errors.Join(err)
+	}
+	return fmt.Errorf("%w: errors: %w", ErrInvalidCertificate, certErrors)
```
fasmat: I think this can be simplified to:

```suggestion
	for _, key := range matchingKeys {
		_, err := shared.VerifyCertificate(certificate, key, nodeID)
		if err == nil {
			return nil
		}
	}
	return ErrInvalidCertificate
```

Reason being that a) all `err` joined to `certErrors` will be the same (some error failing to validate the certificate with the given key) and b) `ErrInvalidCertificate` already implies that the provided certificate failed to validate and adding `certErrors` would just repeat the same error again.

server/server.go
```diff
@@ -221,6 +255,7 @@ func (s *Server) Start(ctx context.Context) error {
 	defer cancel()
 	err = server.Shutdown(shutdownCtx)
 	grpcServer.GracefulStop()
+	configGrpcServer.GracefulStop()
 	return errors.Join(err, serverGroup.Wait())
```
fasmat: Instead of joining these errors here why not just log them when they happen? i.e.:

```go
if err := server.Shutdown(shutdownCtx); err != nil {
    logger.Sugar().Errorf("failed to shutdown server: %s", err)
}
if err := serverGroup.Wait(); err != nil {
    logger.Sugar().Errorf("error when waiting to shutdown server: %s", err)
}
return ctx.Err()
```

The the only thing the caller does is take the errors joined here, wrap them again with more text and then print them in a single log statement to the console. If `ctx.Err()` is used the caller can still check that the function was stopped because of the context being cancelled (rather than an actual error that should be handled or logged) and then continue to exit. Wdyt?

fasmat (commented):
Should we fix the Docker linting error as well in this PR so GH stops complaining that `AS` should be capitalized when `FROM` is? 🙂

registration/registration.go
poszu: I'm lost, what issue do you mean? Frankly, I think either case is fine and makes little (zero) difference in reality (reloading keys is supposed to be rare).

poszu:
> > Which tests depend on the order they are executed in?
> 
> `"valid certificate (with loaded trusted keys)"` if executed individually does not read key files from disk:
> 
> ```shell
> go test -run ^Test_CheckCertificate$/^valid_certificate_\(with_loaded_trusted_keys\)$ github.com/spacemeshos/poet/registration
> ```
> 
> but it actually looks like if executed like this the test isn't run at all. Maybe that's because the tests are double nested?

I've already removed this test and refactored others. Please check again :)

registration/registration.go
fasmat: I mean in the case where we wait here (other routines are currently reading or writing to the `map`) we want to wait before we read from disk to avoid putting more load on the system or not? Sure this only realistically happens if someone reloads the certificates while a lot of nodes are submitting, but maybe that's not that rare.

E.g. the operator forgot to add a trusted certificate and realizes this while users are submitting and tries to fix it. In that case we want to block submissions before reading the certificates from disks or not?

server/server.go
poszu: Makes sense. I updated the code but changed the last line to return `nil`. Cancelling the context is the way to stop the server (I mean: this is a normal condition - hence `nil`).

fasmat (approved):

PR #522, opened 2024-10-02
Update used go version to 1.23.2

fasmat:


poszu (approved):
Could you also update the Dockerfile?

PR #535, opened 2024-11-06
Update merkle-tree library

fasmat:
This integrates the updated library from https://github.com/spacemeshos/merkle-tree/pull/51 into PoET.

I added a check that any submitted challenge is 32 bytes. The check is necessary to do early because the merkle tree is only built when the round is closed and will now fail if any leafs are included that aren't 32 bytes long.


noamnelke (approved):
👍🏻

